PR #45, opened 2014-10-08
amd + umd support

mente:
Hello!

This PR adds basic support for AMD and UMD that is already in video.js and missing in this plugin.
I've slightly updated jshint config file to pass tests.


gkatsev:
Merged in via #72 and published as `v1.1.0`.

PR #51, opened 2014-11-14
Fixes poster flashing in between prerolls and content playing.

bclwhitaker:
Set display to none on the poster and play button elements while preroll handshake is in progress.


dmlap:
@bclwhitaker ping

bclwhitaker:
Closing.  We've addressed this further up the stack.

zang:
@bclwhitaker  I got the same problem recently. Could you please point me to the solution if it's been solved? thanks

PR #56, opened 2014-12-09
Add blanket.js for code coverage.

bclwhitaker:
Update tests so that they don't throw when run in a browser.

### DIFF
--- a/test/videojs.ads.test.js
+++ b/test/videojs.ads.test.js
@@ -87,16 +87,21 @@ test('pauses to wait for prerolls when the plugin loads after play', function() 
 });
 
 test('stops canceling play events when an ad is playing', function() {
-  var callback;
+  var
+    clearImmediateCallCount = 0, 
+    callback;
 
-  expect(3);
+  expect(4);
   // capture setImmediate callbacks to manipulate invocation order
   window.setImmediate = function(cb) {
     callback = cb;
     return 1;
   };
-  window.clearImmediate = function() {
+  window.clearImmediate = function(id) {
     callback = null;
+    if (id === player.ads.cancelPlayTimeout) {
+      clearImmediateCallCount++;
+    }
   };
 
   player.trigger('play');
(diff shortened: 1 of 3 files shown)


test/videojs.ads.test.js
```diff
@@ -709,17 +708,14 @@ test('adscanceled allows us to transition from content-set to content-playback',
 
 test('adscanceled allows us to transition from ads-ready? to content-playback', function() {
   var callback;
-  expect(6);
+  expect(5);
```
dmlap: @bclwhitaker do you know the answer to gkatsev's question?

bclwhitaker: window.clearImmediate was being called twice causing the assert inside of it to be executed twice. When running with the code coverage tool QUnit was throwing errors because of the asserts within these timeout overrides. I moved the assert out of the functions and therefore only needed to be one.  I can add back a check that clearImmediate is actually being called with player.ads.cancelPlayTimeout to more closely resemble the initial, but there will still be one less assert.

bclwhitaker: @dmlap @gkatsev Upon further review, I removed the previous test updates I had made alltogether.  You will notice that if you run the tests on master in a browser now they will throw with this error:

```
Uncaught Error: assertion outside test context, was     at window.clearImmediate (http://localhost:8081/test/videojs.ads.test.js:720:21)
    at clearImmediate (http://localhost:8081/src/videojs.ads.js:104:33)
    at Object.fsm.preroll?.leave (http://localhost:8081/src/videojs.ads.js:363:15)
    at http://localhost:8081/src/videojs.ads.js:477:38
    at fsmHandler (http://localhost:8081/src/videojs.ads.js:485:9)
    at HTMLDivElement.e (http://localhost:8081/node_modules/video.js/dist/video-js/video.js:13:254)
    at HTMLDivElement.e.Y.e.Y (http://localhost:8081/node_modules/video.js/dist/video-js/video.js:6:286)
Source: 	
http://localhost:8081/libs/qunit/qunit.js:899
```
 The root cause of this was that clearTimeout was not restored in the AdFramework teardown even though its overridden in the tests.  My last commit fixes this and should make this pull a bit more tolerable.

dmlap:
@bclwhitaker any chance you'll have time to update this soon?

PR #57, opened 2014-12-31
Remove restore tracks

gkatsev:
This supersedes and builds upon #47.


src/videojs.ads.js
```diff
@@ -148,8 +154,18 @@ var
...
     if (tech) {
       snapshot.nativePoster = tech.poster;
+
+      suppressedTracks.elements = Array.prototype.slice.call(tech.querySelectorAll('track'));
     }
 
+    i = tracks.length;
+    while (i--) {
+      track = tracks[i];
+      suppressedTracks.list.addTrack_(track);
```
dmlap: Doesn't seem quite right to be using a "private" method here.

src/videojs.ads.js
```diff
@@ -480,28 +533,24 @@ var
       'adstart',  // startLinearAdMode()
       'adend'     // endLinearAdMode()
     ]), fsmHandler);
-    
-    // keep track of the current content source
-    // if you want to change the src of the video without triggering
-    // the ad workflow to restart, you can update this variable before
-    // modifying the player's source
-    player.ads.contentSrc = player.currentSrc();
-
```
dmlap: Was this removal and the changes below intentional?

gkatsev: No, that wasn't intentional.

src/videojs.ads.js
gkatsev: I'm not even sure that `list` here needs to be a `TextTrackList`.

src/videojs.ads.js
```diff
@@ -246,6 +268,37 @@ var
       // just resume playback at the current time.
       player.play();
     }
+
+    i = suppressedTracks.list.length;
+    if (tracks.addTrack_) {
+      while (i--) {
+        track = suppressedTracks.list[i];
+        tracks.addTrack_(track);
```
gkatsev: I debated using `.addTrack_` here a lot. One of the issues with not using it is that currently we don't store the associated src on the TextTrack objects themselves but if we did and we recreated the tracks, we would need to make another request (though, it should be cached) and parse the vtt file again, so, I thought just stuffing it back in was reasonable, if not great.
The main reason that `addTrack_` is pseudo-private is that we need a way to add items to the list internally but it isn't part of the official API spec. Maybe, we should just make them official APIs?

dmlap: How about snapshotting the cues? That avoids the double request/parse and their are public APIs for recreating them.

gkatsev: The issue with that is that tracks added with `addTextTrack` in native mode cannot be removed, so, we'd be stuck with just changing the mode of text tracks' mode to be 'disabled', in which case, we don't need to snapshot anything except the mode's of the tracks.

gkatsev: we've decided to forgot this altogether and just snapshot the modes and set them all to 'disabled'.

PR #58, opened 2015-01-06
Prevent a simulated play event from being triggered when there is no src...

bclwhitaker:
... set and the flash tech is in use.


dmlap:
We shouldn't be writing defensive code for other video.js projects. Let's just fix this through https://github.com/videojs/video-js-swf/issues/136

PR #60, opened 2015-01-13
Feature/content playback event

bclwhitaker:
### This commit would require a major rev as it removes the ad-timeout-playback state.

* Add handling for 'adserror' events from ad integrations.
* All "ad exiting" events (currently adtimeout, adscanceled, adserror, or adend) now change the state of the contrib-ads plugin to 'content-playback'.
* A new property on the fsm, 'triggerevent' tracks the triggering event for all state changes.
* A new event 'contentplayback' fires whenever the content-playback state is entered. The 'triggerevent' property is available on this event to distinguish the cause.
* Removed the ad-timeout-playback state.  This is now functionally equivalent to being in 'content-playback' state and the triggerevent being 'adtimeout'.

```javascript
player.on('contentplayback', function(event) {
  if (event.triggerevent === 'adtimeout') {
    //we would have been in ad-timeout-playback state previously
  }
}
```


README.md
```diff
@@ -39,12 +39,14 @@ Once you call `player.ads()` to initialize the plugin, it provides six interacti
...
- * `readyforpreroll` (EVENT) — Fires the when a content video is about to play for the first time, so your integration can indicate that it wants to play a preroll.
+ * `readyforpreroll` (EVENT) — Fires when a content video is about to play for the first time, so your integration can indicate that it wants to play a preroll.
+ * `contentplayback` (EVENT) - Fires whenever the state changes to content-playback state.  Included in this event is a 'triggerevent' property which indicates what event type triggered the state change. 
 
 And here are the interaction points you use to send information to the ads plugin:
 
  * `adsready` (EVENT) — Trigger this event after to signal that your integration is ready to play ads.
  * `adscanceled` (EVENT) — Trigger this event after starting up the player or setting a new video to skip ads entirely. This event is optional; if you always plan on displaying ads, you don't need to worry about triggering it.
+ * `adserror` (EVENT) - Trigger this event to indicate that an error in the ad integration has ocurred and the the preroll ad flow should abort so that content can resume.
```
dmlap: Shouldn't ad flow abort during midrolls as well?

dmlap: If not, this event seems really close to `adscanceled`. Why not just extend the behavior of that?

src/videojs.ads.js
```diff
@@ -378,22 +382,15 @@ var
...
-
               // signal to ad plugin that it's their opportunity to play a preroll
               player.trigger('readyforpreroll');
-
             },
             leave: function() {
               window.clearTimeout(player.ads.timeout);
-
-              clearImmediate(player.ads.cancelPlayTimeout);
-              player.ads.cancelPlayTimeout = null;
```
dmlap: Why was this removed?

src/videojs.ads.js
```diff
@@ -462,9 +443,11 @@ var
             enter: function() {
               // capture current player state snapshot (playing, currentTime, src)
               this.snapshot = getPlayerSnapshot(player);
-
               // remove the poster so it doesn't flash between videos
               removeNativePoster(player);
+              // We no longer need to supress play events once an ad is playing
```
dmlap: What about when two video elements are in use?

README.md
bclwhitaker: I think its important to distinguish an error from a canceling/skip event for many (especially from an analytics perspective.)

src/videojs.ads.js
```diff
@@ -477,11 +460,33 @@ var
...
+                }
+              }
+            },
             events: {
+              // in the case of a timeout, adsready might come in late.
+              'adsready': function() {
+                if (player.paused()) {
+                  this.state = 'ads-ready';
+                } else {
+                  this.state = 'preroll?';
```
dmlap: I had thought this would work by leaving the player in `content-playback` and then the ad integration would have to trigger late prerolls as if they were midrolls. Is there a reason this flow is better that I'm not thinking of?

README.md
bclwhitaker: We could ad handling for adserror in adplayback state too to cover that case, but it seemed silly for someone to call startLinearAdMode if there was an error to begin with.  What do you think should I add it?

dmlap: Imagine if the ad configuration contained content that timed out or an unsupported VMAP feature was used.

dmlap:
If `adserror` is sticking around, we should add some tests for it.

src/videojs.ads.js
bclwhitaker: This code was in the ad-timeout-playback state.  I was just porting it over.  I think your suggestion makes it harder for ad-integrations cause then they need to know whether they have missed the loop or not with regards to the timing of them triggering ads-ready.

bclwhitaker:
Fair enough, I thought I had.  Will add those now.

src/videojs.ads.js
dmlap: If you're going to remove two states, you can't "just port it over". Are the assumptions made in those states still valid at this point?

src/videojs.ads.js
bclwhitaker: It was consolidated into the enter method of content-playback.

src/videojs.ads.js
bclwhitaker: Not following.  This is where the code from ad-timeout-playback was moved to (which was invoked after putting it into content-playback state).  I am making the assumption (perhaps incorrectly) that we would never want to supress play events while content is playing.  Is this not true?

src/videojs.ads.js
bclwhitaker: Yes, in my investigation the state of the player, etc would be the same here as it would have been in ad-timeout-playback state and having listening for late arriving adsready calls makes sense the same here as it would have there.

README.md
bclwhitaker: OK, I'll add that and update the diagram.

bclwhitaker: Added.  Readme, tests, and diagram updated to reflect this.

src/videojs.ads.js
dmlap: Nice!

src/videojs.ads.js
dmlap: This is the ad-playback enter handler. It seems to me that this says "do not cancel play attempts while ads are playing". That sounds good if a single video element is getting re-used for ad playback but doesn't sound good if one video is playing ads and the other is playing content. I may be misreading it.

src/videojs.ads.js
dmlap: Ok, thanks for elaborating. I can see how this lines up with the old flow but going backward doesn't seem right to me. What about if we just fired a `readyforpreroll` if `adsready` arrived late?

bclwhitaker: I ended up making this change as you suggested.  This basically means that the ad-integrations are on their own in this scenario though (adsready not triggering until content has already started to play).  I concluded that was reasonable because this scenario is something each integration would probably want to deal with differently anyway.  Some might want to choose to halt the content playback play the preroll and resume,  some might want to play the preroll and restart the content, and others might want to just skip the preroll all-together.

dmlap:
@csinchok @mattosborn @shawnbuso I'd be really curious to hear your opinions on the state simplification if you have any

shawnbuso:
Seems like it would be OK to me, thanks for the heads up.

PR #61, opened 2015-01-23
player.remoteTextTracks -> player.textTracks

mcountis:
Plugin was throwing an error looking for player.remoteTextTracks method. I was unable to locate the method in the videojs library, but upon searching through old commits on this file it looks like the method is a remnant that wasn't corrected.

If I am wrong, I'd like to know where to find the method remoteTextTracks on the player object.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -140,7 +140,7 @@ var
   getPlayerSnapshot = function(player) {
     var
       tech = player.el().querySelector('.vjs-tech'),
-      tracks = player.remoteTextTracks(),
+      tracks = player.textTracks(),
       track,
       i,
       suppressedTracks = [],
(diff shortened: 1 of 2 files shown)


gkatsev:
Hey, thanks for the PR. A mistake on our end. We're in the process of adding new text track functionality to videojs (see videojs/video.js#1749) and we accidentally assumed everyone would be using that branch of videojs.
Instead of just removing the usage of it, we should feature detect it and if it's available, use what's in there and otherwise do nothing.

gkatsev:
remoteTextTracks just ladned in videojs master but we'll make sure to release a version that won't break on older versions of videojs.

gkatsev:
We've published another fix for this as v1.0.1.

PR #62, opened 2015-01-23
Always trigger adend when exiting ad-playback state.

bclwhitaker:
Always trigger 'adend' when exiting ad-playback for consistency.  Update state diagram to reflect this change.


dmlap:
LGTM

PR #63, opened 2015-01-25
Ensure that playback is resumed when in content-playback state if a play event was canceled earlier.

bclwhitaker:
Also remove play() call on timeout since this is now handled in the same way and was causing play() to be invoked twice.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -118,15 +118,21 @@ var
       // another cancellation is already in flight, so do nothing
       return;
     }
-
     player.ads.cancelPlayTimeout = setImmediate(function() {
-
       // deregister the cancel timeout so subsequent cancels are scheduled
       player.ads.cancelPlayTimeout = null;
 
+      // pause playback so ads can be handled.
       if (!player.paused()) {
         player.pause();
       }
+
+      // add a contentplayback handler to resume playback when ads finish.
+      player.on('contentplayback', function() {
+        if (player.paused()) {
+          player.play();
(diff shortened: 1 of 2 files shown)


src/videojs.ads.js
```diff
@@ -403,7 +409,6 @@ var
               },
               'adtimeout': function() {
                 this.state = 'content-playback';
-                player.play();
```
bclwhitaker: This was no longer necessary.  This was caught in the 'starts the content video if there is no preroll' test which checks the number of times play is called.

src/videojs.ads.js
```diff
@@ -467,19 +472,17 @@ var
...
-              player.trigger({
-                type: 'contentplayback',
-                triggerevent: fsm.triggerevent
-              });
-              // Make sure that the cancelPlayTimeout is cleared
+              // make sure that any cancelPlayTimeout is cleared
               if (player.ads.cancelPlayTimeout) {
                 clearImmediate(player.ads.cancelPlayTimeout);
                 player.ads.cancelPlayTimeout = null;
-                // Trigger playback if play was canceled and the player is paused.
```
dmlap: Why not just move this outside of the conditional?

dmlap: @bclwhitaker informed me that it's possible to hit this state without receiving a play request from the user if `adscancelled` or `aderror` is fired from one of the early states like `content-set`

dmlap:
LGTM

PR #64, opened 2015-01-29
Prefix ad-related video events

dmlap:
Transform video events during ad playback to simplify the downstream event API. See https://github.com/videojs/video.js/issues/1825 for more details.


bclwhitaker:
Please be sure to test that engagement is reported properly with the onceux and these changes (with prerolls, midrolls, and postrolls) -- There are some example streams for once in socrates.  I assume you are testing extensively with IMA already. There are a number of negative pre-roll examples in socrates that can offer quick validation that we're still handling errors proprerly.

example/example-integration.js
```diff
@@ -14,22 +14,28 @@
...
       //  - adPlaying - whether a linear ad is currently playing
       //  - prerollPlayed - whether we've played a preroll
       //  - midrollPlayed - whether we've played a midroll
+      //  - postrollPlayed - whether we've played a postroll
       state = {},
-      
+
+      // just like any other video.js plugin, ad integrations can
+      // accept initialization options
+      adServerUrl = (options && options.adServerUrl) || "inventory.json",
```
seniorflexdeveloper: Shouldn't this be (options && options.adServerUrl) ? options.adServerUrl : "inventory.json", or am I just reading wrong?

example/example-integration.js
```diff
@@ -99,29 +107,29 @@
...
+
       if (state.midrollPlayed) {
         return;
       }
-      
+
       var currentTime = player.currentTime(), opportunity;
-      
+
       if ('lastTime' in state) {
```
seniorflexdeveloper: Just curious, what is the 'lastTime' state for?

seniorflexdeveloper:
@gkatsev - couple of semantics questions but functionally LGTM.

example/example-integration.js
dmlap: `&&` returns the second value when it succeeds, which is `options.adServerUrl` in this case. This line is being overly clever.

seniorflexdeveloper: Got it, thx.

example/example-integration.js
gkatsev: I think this makes it play the midroll if you seek.

PR #65, opened 2015-02-05
Fix remoteTextTracks error

shoshomiga:


gkatsev:
Thanks for the PR!

gkatsev:
remoteTextTracks just ladned in videojs master but we'll make sure to release a version that won't break on older versions of videojs.

gkatsev:
We've published another fix for this as v1.0.1.

PR #72, opened 2015-03-12
Umd

gkatsev:
Fixes #45

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -1,7 +1,7 @@
 (function(factory){
     "use strict";
     if (typeof define === 'function' && define.amd) {
-        define(['./video'], function(vjs){factory(window, document, vjs);});
+        define(['videojs'], function(vjs){factory(window, document, vjs);});
     } else if (typeof exports === 'object' && typeof module === 'object') {
         factory(window, document, require('video.js'));
     } else {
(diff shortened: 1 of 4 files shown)


src/videojs.ads.js
```diff
@@ -1,3 +1,13 @@
+(function(factory){
+    "use strict";
+    if (typeof define === 'function' && define.amd) {
+        define(['videojs'], function(vjs){factory(window, document, vjs);});
```
dmlap: won't this register a named module called `videojs`?

gkatsev: No, the array indicates that it's a dependency.

gkatsev: Which reminds me, should we name this in the amd?
```js
define('videojs-contrib-ads', ['videojs'], function(vjs) { factory(window, document, vjs); });
```

gkatsev:
The last break is from the prepublish script. Looking into correct action here.

src/videojs.ads.js
dmlap: Oops. Yeah, it makes sense as a dependency. I'm inclined to go with a named module so we don't run into the concatenation problem in downstream dependencies again.

dmlap:
lgtm

PR #76, opened 2015-03-20
Trigger ended from content-resuming on ended

gkatsev:
Only trigger ended from content-resuming if we don't get a native ended event.
Still requires:
* [x] Tests

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -583,6 +583,17 @@ var
             }
           },
           'content-resuming': {
+            enter: function() {
+              if (this.snapshot.ended) {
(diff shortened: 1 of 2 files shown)


src/videojs.ads.js
```diff
@@ -583,6 +583,17 @@ var
             }
           },
           'content-resuming': {
+            enter: function() {
+              if (this.snapshot.ended) {
+                window.clearTimeout(player.ads._fireEndedTimeout);
+                player.ads._fireEndedTimeout = window.setTimeout(function() {
```
dmlap: Could you add a comment explaining why the timeout is necessary?

gkatsev: Done.

PR #78, opened 2015-03-31
Don't stop propagation on video element events.

### LINKED ISSUE #46: Video does not resume on Android
On Android (tested on Moto X, 4.4.4, Chrome 38) the video will not resume/start after ad playback. This is because `tech.seekable.length` can remain `0` even though a source is set and playable, and `tech.readyState` is `4` (`HAVE_ENOUGH_DATA`).

Is there a situation where `tech.readyState` couldn't be used to test the playable state in [tryToResume()](https://github.com/videojs/videojs-contrib-ads/blob/master/src/videojs.ads.js#L191-210)?

gkatsev:
Previously, event.stopImmediatePropagation was used on redispatched
events but that was causing issues because the actual native events were
getting stopped in addition to the player events. Updated to stop
immediate propagation manually so only the player events will stop.
The changes to stopImmediatePropagation mean that this cannot be used
with the google closure compiled version of videojs.

Also, tryToResume now checks the video element's readyState for trying
to resume. Fixes #46. Android devices in particular fail to have a
seekable property after reloading a video, especially, with HLS
content.

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -5,7 +5,7 @@
   <title>Example Ad Integration</title>
   <!-- Load local video.js -->
   <link rel="stylesheet" href="../node_modules/video.js/dist/video-js/video-js.min.css">
-  <script src="../node_modules/video.js/dist/video-js/video.js"></script>
+  <script src="../node_modules/video.js/dist/video-js/video.dev.js"></script>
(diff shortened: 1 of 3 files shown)


gkatsev:
Approved by @dmlap.

PR #80, opened 2015-04-01
add commonjs enablers

orcaman:
The package is suited to be required as a commonjs module, except that some settings are missing from the package.json file to make this possible. I've added these changes, and was then able to successfully consume it as a commonjs module ("require('videojs-contrib-ads');" from my code).  

Feel free to merge if you find it useful, I guess it would benefit people working with this commonjs style.


incompl:
This project has been updated to follow the package format dictated by [generator-videojs-plugin](https://github.com/videojs/generator-videojs-plugin) which I believe supports commonjs.

PR #83, opened 2015-04-09
UMD

gkatsev:
This change involves putting the main file back into `videojs.ads.global.js` without any UMD, it is then wrapped in UMD and put into `videojs.ads.js`.
When developing, you should develop with `videojs.ads.global.js` and then have `grunt umd` generate the UMD file for us.


src/videojs.ads.js
```diff
@@ -1,13 +1,19 @@
...
-    "use strict";
-    if (typeof define === 'function' && define.amd) {
-        define('videojs-contrib-ads', ['videojs'], function(vjs){factory(window, document, vjs);});
-    } else if (typeof exports === 'object' && typeof module === 'object') {
-        factory(window, document, require('video.js'));
-    } else {
-        factory(window, document, videojs);
-    }
-})
+(function (root, factory) {
```
gkatsev: even though git thinks it's different, this file is actually our generated file (it has a new header) and the other file is what this file used to be.

apadhye:
:+1:

PR #84, opened 2015-04-14
Text tracks on ads

lemoniasty:
This path hides the text tracks while an ad is being played and restores when finished.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -567,6 +567,11 @@ var
               // capture current player state snapshot (playing, currentTime, src)
               this.snapshot = getPlayerSnapshot(player);
 
+               //Hide text track while the ad is being played.
+                if (player.textTracks().length > 0) {
+                player.textTrackDisplay.hide();
+              }
+
               // remove the poster so it doesn't flash between videos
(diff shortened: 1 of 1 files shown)


gkatsev:
Looks good, though, not sure about the use case. We currently try to remove and disable text tracks when ads are playing and then restoring the tracks and modes after they end, so, theoretically no text tracks should be available.
Can you elaborate on your use case and any issues, if you're seeing them.

lemoniasty:
Hi Gary,
that's the functionality I was expecting after flicking through the changelog but unfortunately it doesn't work. I've sent you two links - first were the problem is visible (text tracks displayed on a preroll) and second with the patch applied (no text tracks on a preroll but displayed on the movie).

Link with the problem: http://infoterra.civ.pl/devel/player_use_case/examples/
Link with the patch applied: http://infoterra.civ.pl/devel/player_use_case/examples/index2.html

I hope it's more clear now :)

lemoniasty:
Hi Gary,
Is everything clear for you?

incompl:
This is over a year old and it happened before I was involved in the project, so just want to check in. Is this still a desired change?

PR #91, opened 2015-05-20
Properly clean up the video player between tests

imbcmdth:
Not much else to add.. @gkatsev can you think of a better description?


gkatsev:
"dispose player between runs on events test page"

dmlap:
LGTM. Waiting for Travis.

gkatsev:
Not that we shouldn't wait for travis, but we don't run this test page on the CI: https://github.com/videojs/videojs-contrib-ads/blob/master/Gruntfile.js#L74-L76

PR #92, opened 2015-06-11
Added dist folder to the project

carpasse:
The 'dist' folder is missing in the repo and it forces the users of the plugin to download the plugin and build it to be able to use the minified versions. 

This PR adds the dist folder to the repo so that we can used the minified version of the plugin or he umd version whenever we download the plugin from bower or npm.


incompl:
This project has been updated to follow the conventions established by [generator-videojs-plugin](https://github.com/videojs/generator-videojs-plugin): `dist` is not checked into the project, but if you install with npm, the built files in the dist folder are present.

PR #100, opened 2015-07-24
Adstart adend checks

bclwhitaker:
Use the ad-playback state to make adstart and adend trigger more accurately. Remove duplicate code around triggering adstart state that is already included in the ad-playback.enter().


bcvio:
LGTM

gkatsev:
LGTM. Commit history should be cleaned up before or during the merge.

dmlap:
lgtm

PR #101, opened 2015-07-24
Add skipLinearAdMode and 'adskip' event handling.

bclwhitaker:
Having a skipLinearAdMode method which triggers an 'adskip' event will allow plugin developers to quickly skip the preroll flow and move directly to content-playback when they:
* Already know they don't have a preroll to play
* There are no preroll ads returned in the ad response
* There are only non-linear ads returned in the ad response.

RIght now, some plugins work around this deficiency by triggering adserror or adstimeout events.  This allows for a more semantically correct transition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ And here are the interaction points you use to send information to the ads plugi
  * `adserror` (EVENT) - Trigger this event to indicate that an error in the ad integration has ocurred and any ad states should abort so that content can resume.
  * `ads.startLinearAdMode()` (METHOD) — Call this method to signal that your integration is about to play a linear ad. This method triggers `adstart` to be emitted by the player.
  * `ads.endLinearAdMode()` (METHOD) — Call this method to signal that your integration is finished playing linear ads, ready for content video to resume. This method triggers `adend` to be emitted by the player.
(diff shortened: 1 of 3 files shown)


gkatsev:
LGTM, but the commit history looks pretty messy. Would be nice to clean it up before or during the merge.

README.md
```diff
@@ -49,6 +49,8 @@ And here are the interaction points you use to send information to the ads plugi
  * `adserror` (EVENT) - Trigger this event to indicate that an error in the ad integration has ocurred and any ad states should abort so that content can resume.
  * `ads.startLinearAdMode()` (METHOD) — Call this method to signal that your integration is about to play a linear ad. This method triggers `adstart` to be emitted by the player.
  * `ads.endLinearAdMode()` (METHOD) — Call this method to signal that your integration is finished playing linear ads, ready for content video to resume. This method triggers `adend` to be emitted by the player.
+ * `ads.skipLinearAdMode()` (METHOD) — Call this method to signal that your integration is not going to play a linear ad and should skip directly to content playback (no linear ads in response, already know there is no preroll to wait for). This method triggers `adskip` to be emitted by the player.  You should always use `endLinearAdMode()` to exit the linear `ad-playback` state.  For this reason, `adskip` will not be triggered and has no effect if you are already in the `ad-playback` state.
```
dmlap: I don't quite follow why the note at the end about endLinearAdMode() is important. Also, the examples in the parenthetical of the first sentence are great but brevity may have won about a bit too much over clarity. What about a "For example, ..." or "e.g." and ", *you* already know there is no preroll to wait for".

dmlap:
Just a minor suggestion on documentation. Otherwise, lgtm

README.md
bclwhitaker: I was worried that people might think this method was somehow related to the typical advertising SDK concept of skipping an already playing ad.  It sounds like I may have made it more confusing with my comments though. I'll remove them.

PR #107, opened 2015-08-17
While we're playing ads, disable loading of tracks

gkatsev:


gkatsev:
This probably does need some tests. And maybe a simplification of how we're disabling the tracks. We're almost at a belt-and-suspenders level of disabling.

src/videojs.ads.js
```diff
@@ -155,11 +160,43 @@ var
...
+      while (i--) {
+        track = trackEls[i].track;
+
+        for (j = 0; j < suppressedTracks.length; j++) {
+          if (suppressedTracks[j].track === track) {
+            suppressedTracks[j].mode = track.mode;
+          }
+        }
+
+        track.mode = 'disabled';
```
dmlap: Two questions: 

I don't think `getPlayerSnapshot()` should be making side-effecting changes to the player (or in this case, setting event listeners that create side-effecting changes). Can this be moved into the transitions into and out of linear ad mode?

This loop seems to unconditionally disable the track, whether it is on the suppressed list or not. Is that correct? Should it be?

dmlap:
Agreed on adding some test cases.

src/videojs.ads.js
gkatsev: We do currently set the remote text tracks to disabled anyway. Enabling and disabling this in the transitions in and out of linear ad mode seems fine.

We probably should add any tracks that aren't already on the `suppressedTracks` list into that list for restoration. The `suppressedTracks` list is originally made from remote text tracks but it isn't necessarily the case that all the tracks would be available in that list that we want to disable. We wouldn't want any cues to fire while an ad is playing. I've been thinking that maybe this function is how we should be handling all the text tracks: loop through all the tracks available from `textTracks()` and also anything we can get via the `querySelectorAll('track')` (since not everything will necessarily show up in `textTracks()` until vjs5) and update the snapshot on every `adplaying`.
Though, maybe that refactor is a bit too ambitious for this PR/fix.

incompl:
This is over a year old and it happened before I was involved in the project, so just want to check in. Is this still a desired change? Any plans to rebase it and add tests?

gkatsev:
Um... yeah, we potentially want to get this merged in but wow is this old. Totally forgot that it existed.
I guess maybe we should evaluate how many issues related to captions during ads we have to see whether we need this soon because we can always update it at a later date.

incompl:
This is so old at this point without activity that I'll close and we can reopen if we want to revisit this.

PR #109, opened 2015-08-18
Update README to utilize `adended` event. Removed unnecessary `durati…

chemoish:
…onchange` event (However, supposed to be `addurationchange`). Apparently some formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Here are the events that communicate information to your integration from the ad
 
  * `contentupdate` (EVENT) — Fires when a new content video has been assigned to the player, so your integration can update its ad inventory. _NOTE: This will NOT fire while your ad integration is playing a linear Ad._
  * `readyforpreroll` (EVENT) — Fires when a content video is about to play for the first time, so your integration can indicate that it wants to play a preroll.
- * `contentplayback` (EVENT) - Fires whenever the state changes to content-playback state.  Included in this event is a 'triggerevent' property which indicates what event type triggered the state change. 
+ * `contentplayback` (EVENT) - Fires whenever the state changes to content-playback state.  Included in this event is a 'triggerevent' property which indicates what event type triggered the state change.
 
 And here are the interaction points you use to send information to the ads plugin:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,29 +64,27 @@ This is not actually a runnable example, as it needs more information as specifi
...
     // play your linear ad content
     player.src('http://url/to/your/ad.content');
-    player.one('durationchange', function(){
-      player.play();
-    });
-    // when all your linear ads have finished...
-    player.one('ended', function() {
+
+    // when all your linear ads have finished… do not confuse this with `ended`
+    player.one('adended', function() {
```
gkatsev: `adended` is when any ad finished. We want `adend` here. I guess we forgot to update this example during the 2.0 update.

chemoish: Hmm... ok, it looks like `adended` gets called before `adend`, though in this implementation, it would seem that the `one()` call makes it do exactly the same thing?

Will update.

chemoish:
Also note:

```js
// tell videojs to load the ad
var media = state.inventory[Math.floor(Math.random() * state.inventory.length)];
player.src(media);

// when it's finished
player.one('adended', function() {
  // play your linear ad content, then when it's finished ...
  player.ads.endLinearAdMode();
  state.adPlaying = false;
});
```

https://github.com/videojs/videojs-contrib-ads/blob/master/example/example-integration.js#L78

README.md
chemoish: Also note:

```js
// tell videojs to load the ad
var media = state.inventory[Math.floor(Math.random() * state.inventory.length)];
player.src(media);

// when it's finished
player.one('adended', function() {
  // play your linear ad content, then when it's finished ...
  player.ads.endLinearAdMode();
  state.adPlaying = false;
});
```

https://github.com/videojs/videojs-contrib-ads/blob/master/example/example-integration.js#L78

chemoish: Actually, I believe its `adended` still as… `endLinearAdMode()` is what triggers `adend`.

gkatsev: Oh, I see. Yeah, it should be `adended`. I misread what this example was trying to do.
Basically, it'll end the ad block as soon as the first ad has finished playing.

chemoish: :successkid:

PR #110, opened 2015-08-18
Update current content video to be sourced from `snapshot.source` ins…

chemoish:
…tead of `snapshot.src` and `snapshot.type`.

Remove `snapshot.type` since it's unnecessary. I also suppose there was an existing issue where the `type` did not exist, but I am guessing that it was a silent error.

This will no longer be a problem as the current content video will be initialized with the exact same data that it was original configured with.

Minor formatting.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -151,10 +151,10 @@ var
       i,
       suppressedTracks = [],
       snapshot = {
-        ended: player.ended(),
-        src: player.currentSrc(),
         currentTime: player.currentTime(),
-        type: player.currentType()
+        ended: player.ended(),
+        source: player.currentSource(),
+        src: player.currentSrc()
       };
 
     if (tech) {
@@ -313,7 +313,7 @@ var
       player.one('contentloadedmetadata', restoreTracks);
 
       // if the src changed for ad playback, reset it
-      player.src({ src: snapshot.src, type: snapshot.type });
+      player.src(snapshot.source);
       // safari requires a call to `load` to pick up a changed source
       player.load();
       // and then resume from the snapshots time once the original src has loaded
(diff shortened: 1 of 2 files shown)


chemoish:
Blocked by https://github.com/videojs/video.js/pull/2479

src/videojs.ads.js
```diff
@@ -151,10 +151,10 @@ var
       i,
       suppressedTracks = [],
       snapshot = {
-        ended: player.ended(),
-        src: player.currentSrc(),
         currentTime: player.currentTime(),
-        type: player.currentType()
+        ended: player.ended(),
+        source: player.currentSource(),
+        src: player.currentSrc()
```
dmlap: Do we need `src` anymore?

chemoish: Yes, `src` is used to validate the ad against the current content.

https://github.com/chemoish/videojs-contrib-ads/blob/feature/currentSource/src/videojs.ads.js#L303
https://github.com/chemoish/videojs-contrib-ads/blob/feature/currentSource/src/videojs.ads.js#L308
https://github.com/chemoish/videojs-contrib-ads/blob/feature/currentSource/src/videojs.ads.js#L413
https://github.com/chemoish/videojs-contrib-ads/blob/feature/currentSource/src/videojs.ads.js#L451

`type` on the other hand was only used to regenerate the `src()` call, so I removed it.

incompl:
This is over a year old and it happened before I was involved in the project, so just want to check in. Is this still a desired change?

chemoish:
I personally don't need this anymore. We can close it—I originally needed it for preroll (which i don't have anymore).

For all complex sources, DRM content specifically, you need more than just the `src` and the `type` when swapping sources.

This applies here, chromecast, and whatever other things leverage complex sources.

PR #111, opened 2015-08-21
Set has-started when an ad starts playing

### LINKED ISSUE #93: .vjs-has-started class not being added to the player when the ad plays
Hi guys

I have migrated from 1.1.0 to 2.1.0 and one of the biggest issues that I found was that the class 
'vjs-has-started' was not added to player when the ad was playing. 

This prevented the poster from disapearing when the ad was playing and the control bar from appearing and disappearing (when the user was active or inactive). 

In the end I had to ad the following styles to my css

```
.vjs-ad-playing .vjs-poster {
  display: none;
}

/* Show the control bar only once the video has started playing */
.vjs-ad-playing .vjs-control-bar {
  display: block;
  /* Visibility needed to make sure things hide in older browsers too. */

  visibility: visible;
  opacity: 1;
  /* transition */
  -webkit-transition: visibility 0.1s, opacity 0.1s;
  -moz-transition: visibility 0.1s, opacity 0.1s;
  -o-transition: visibility 0.1s, opacity 0.1s;
  transition: visibility 0.1s, opacity 0.1s;
}

/* Hide the control bar when the video is playing and the user is inactive  */
.vjs-ad-playing.vjs-user-inactive.vjs-playing .vjs-control-bar {
  display: block;
  visibility: hidden;
  opacity: 0;
  /* transition */
  -webkit-transition: visibility 1s, opacity 1s;
  -moz-transition: visibility 1s, opacity 1s;
  -o-transition: visibility 1s, opacity 1s;
  transition: visibility 1s, opacity 1s;
}
```

I think we should add the previous styles to the plugin (they are pretty much copy pasted from vjs default skin) or investigate why the class 'vjs-has-started' is not being added to the player.

imbcmdth:
Fixes #93.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -401,6 +401,10 @@ var
           });
         },
         redispatch = function(event) {
+          if (event.type === 'play') {
+            player.hasStarted(true);
+          }
+
           if (player.ads.state === 'ad-playback') {
             triggerEvent('ad', event);
 


src/videojs.ads.js
```diff
@@ -401,6 +401,10 @@ var
           });
         },
         redispatch = function(event) {
+          if (event.type === 'play') {
+            player.hasStarted(true);
```
dmlap: This seems like a weird place to fix this. Why was it occurring?

imbcmdth:
This issue was fixed in v4.12.0.

PR #112, opened 2015-08-24
Test that the vjs-has-started class is added once a preroll begins

imbcmdth:
This test is to ensure that we do not get a regression in video.js that results in the `vjs-has-started` class not being applied during preroll ad playback.

### DIFF
--- a/test/videojs.ads.test.js
+++ b/test/videojs.ads.test.js
@@ -132,6 +132,15 @@ test('adstart is fired before a preroll', function() {
   equal(1, adStarts, 'a preroll triggers adstart');
 });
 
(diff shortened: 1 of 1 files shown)


bclwhitaker:
Some would argue that the [tests for this in video.js](https://github.com/videojs/video.js/blob/ecca3a1b5866eeae651588ee24a66d22c279dc13/test/unit/player.test.js#L499-L513) should be sufficient.  But since this class in [managed a little for postroll handling](https://github.com/videojs/videojs-contrib-ads/blob/51e7f832b3459825d96e6bdf4c476de29eeaa2bf/src/videojs.ads.js#L421-L429) I think its fine to add here too.

PR #113, opened 2015-08-28
Prefix videojs events in addition to the html5 spec events.

bclwhitaker:


src/videojs.ads.js
```diff
@@ -437,6 +434,11 @@ var
           }
         };
 
+      //Add video.js specific events
+      videoEvents.push('firstplay');
+      videoEvents.push('loadedalldata');
```
dmlap: Personally, I wish we just dropped these events from video.js but that's not a blocker for this PR.

@videojs/core-committers do we still feel like these events are important?

dmlap:
lgtm

PR #116, opened 2015-09-08
VideoJS 5.0 Updates (DO NOT MERGE)

misteroneill:
This updates the plugin for use with VideoJS 5.0 changes.

There is one functional change that should not affect consumers of the plugin whatsoever: there is no longer any use of `setImmediate` (non-standard) or `requestAnimationFrame`; rather, this is simplified to use `setTimeout` universally. This not only makes code easier to reason about - no need to think about the minor differences in 3 separate APIs - but it makes testing using Sinon's fake timers more straight-forward.

## Testing Changes

The bulk of the changes in this PR are related to testing; this seemed like a good opportunity for that work. There were significant sources of fragility in the tests previously and this sweeps the entire test suite with changes like:

### Reduced reliance on globals

Each test should have whatever scaffolding it needs attached to `this`, imported from other sources, or hidden in a closure. Leaking globals is a good way to cause hard-to-debug issues.

### Eliminated duplicated code

Particularly in setup/teardown processes, there was a ton of repetition. This should now be minimized via `sharedModuleHooks`.

### Switched to QUnit 2.0-style tests

- The only QUnit global that should be used now is `QUnit`.
- `QUnit.module` should be passed `beforeEach`/`afterEach` hooks instead of the deprecated `setup`/`teardown`.
- All assertion functions can be referenced on the `assert` object passed into each test (defined by `QUnit.test`).

### Reduced assertion fragility

There was a lot of use of `QUnit.assert.equal`, which could potentially lead to bugs. `QUnit.assert.strictEqual` is used now, for the same reason `===` is preferred to `==`.

### Used Sinon extensively 

Fake timers were already in use in a few places, but this makes that universal (particularly in the Flash tests). All custom/manual function spying is replaced by uses of `sinon.spy` - no more manual counting or `push`ing onto arrays!


.gitignore
```diff
@@ -1,2 +1,3 @@
+/dist/
```
gkatsev: do we have an npmignore? Because otherwise, npm would ignore the dist folder as well.

misteroneill: Good point. Added an empty .npmignore file.

gkatsev: Also, when this is ready to go, we should try publishing it to npm under a next or beta tag.

misteroneill:
This can probably be reviewed, but should not be merged until we are sure it should be in `master` since it's not 4.x compatible!

misteroneill:
Also, worth noting: two tests are currently broken in `events.html`, but this is a result of [a bug](https://github.com/videojs/video.js/issues/2588) in VideoJS.

package.json
```diff
@@ -30,7 +30,9 @@
     "grunt-contrib-watch": "~0.4.0",
     "grunt-umd": "^2.3.3",
     "load-grunt-tasks": "^3.1.0",
+    "lodash": "~3.10.1",
     "qunitjs": "^1.17.1",
-    "video.js": "^4.12.4"
+    "sinon": "~1.16.1",
```
dmlap: Nitpick but isn't `^` the best-practice version specifier these days?

src/videojs.ads.js
```diff
@@ -234,13 +226,20 @@ var
...
             if (!ended) {
               player.play();
             }
             player.off('ended', updateEnded);
+            player.ads.resumeEndedTimeout = null;
           }, 250);
           player.on('ended', updateEnded);
+
+          // Need to clear the resume/ended timeout on dispose. If it fires
+          // after a player is disposed, an error will be thrown!
```
dmlap: This is a bug that would exist in the current release, right?

test/videojs.ads.snapshot.js
```diff
@@ -1,467 +1,365 @@
...
+  this.player.src('//example.com/ad.mp4');
+  this.player.trigger('loadstart');
+  this.player.trigger('loadedmetadata');
+  this.player.ads.endLinearAdMode();
+
+  // resume playback
+  this.player.src('http://media.w3.org/2010/05/sintel/trailer.mp4');
+  this.player.trigger('loadstart');
+  this.player.trigger('canplay');
+  this.clock.tick(1000);
```
dmlap: Why is this delay now necessary?

test/videojs.ads.snapshot.js
```diff
@@ -1,467 +1,365 @@
...
+  assert.strictEqual(this.contentPlaybackReason(1), 'ended', 'The reason for content-playback should have been ended');
+});
 
-  player.load = function() {
-    loadCalled = true;
-  };
+QUnit.test('snapshot does not resume playback after a burned-in post-roll', function(assert) {
+  assert.expect(9);
+  this.player.trigger('adsready');
+  this.player.tech.trigger('play');
```
dmlap: Any reason why we're firing this on the tech and not the player now?

test/videojs.ads.snapshot.js
```diff
@@ -1,467 +1,365 @@
...
-  player.play = function() {
-    playCalled = true;
-  };
-  i = callbacks.length;
-  while (i--) {
-    callbacks[i]();
-  }
-  ok(playCalled, 'called play() to trigger an "ended"');
+  var playSpy = sinon.spy(this.player, 'play');
+  this.clock.tick(10000);
```
dmlap: Shouldn't this be more like 250?

test/videojs.ads.snapshot.js
```diff
@@ -495,38 +391,39 @@ test('When captions are enabled, the video\'s tracks will be disabled during the
...
     src: 'http://example.com/movie.mp4',
     type: 'video/mp4'
   });
 
-  player.on('readyforpreroll', function() {
-    player.ads.startLinearAdMode();
-  });
-  player.trigger('adsready');
-  player.trigger('play');
+  this.player.on('readyforpreroll', videojs.bind(this, function() {
```
dmlap: `player.on` automatically binds `this` to `player` for you.

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
-  player.trigger('play');
-  player.trigger('play');
-
-  equal(2, pauses, 'play attempts are paused');
-});
+QUnit.test('pauses to wait for prerolls when the plugin loads before play', function(assert) {
+  assert.expect(1);
+  var spy = sinon.spy(this.player, 'pause');
 
-test('pauses to wait for prerolls when the plugin loads after play', function() {
```
dmlap: Where did this test go?

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
+  assert.strictEqual(this.contentPlaybackSpy.callCount, 1, 'A content-playback event should have triggered');
+  assert.strictEqual(this.contentPlaybackReason(), 'adtimeout', 'The reason for content-playback should have been adtimeout');
+});
+
+QUnit.test('starts the content video if there is no preroll', function(assert) {
+  assert.expect(3);
+  var spy = sinon.spy(this.player, 'play');
+  this.player.trigger('adsready');
+  this.player.tech.trigger('play');
+  this.clock.tick(1);
```
dmlap: Does this tick mean it is no longer ok to trigger `adtimeout` in a play event handler?

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
 
-  player.trigger('adsready');
-  player.trigger('play');
-  player.trigger('adtimeout');
-  player.trigger('ended');
-  player.trigger('adtimeout');
-  player.ads.snapshot.ended = true;
-  player.trigger('ended');
+  this.player.trigger('adsready');
+  this.player.tech.trigger('play');
```
dmlap: I'm curious again why this moved to the tech.

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
+  assert.strictEqual(this.player.ads.state, 'preroll?', 'the player is waiting for prerolls');
+  this.player.tech.trigger('play');
+  assert.ok(this.clock.timers.hasOwnProperty(String(this.player.ads.cancelPlayTimeout)), 'a single cancel-play is registered');
+  this.player.tech.trigger('play');
+  this.player.tech.trigger('play');
+  this.player.tech.trigger('play');
+  assert.ok(this.clock.timers.hasOwnProperty(String(this.player.ads.cancelPlayTimeout)), 'a single cancel-play is registered');
+});
```
dmlap: It seems like this is doing an existence check, not verifying that a callback wasn't registered multiple times. Am I missing something?

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
+  assert.strictEqual(this.contentPlaybackReason(), 'adskip', 'The reason for content-playback should have been adskip');
+});
+
+QUnit.test('adserror in postroll? transitions to content-playback and fires ended', function(assert) {
+  assert.expect(8);
+  assert.strictEqual(this.player.ads.state, 'content-set');
+  this.player.trigger('adsready');
+  assert.strictEqual(this.player.ads.state, 'ads-ready');
+  this.player.tech.trigger('play');
+  this.player.trigger('adtimeout');
```
dmlap: A `setImmediate` callback was previously required here. Any idea why it's no longer necessary?

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
+  assert.strictEqual(this.player.ads.state, 'content-playback');
+});
+
+QUnit.test('adtimeout in postroll? transitions to content-playback and fires ended', function(assert) {
+  assert.expect(8);
+  assert.strictEqual(this.player.ads.state, 'content-set');
+  this.player.trigger('adsready');
+  assert.strictEqual(this.player.ads.state, 'ads-ready');
+  this.player.tech.trigger('play');
+  this.player.trigger('adtimeout');
```
dmlap: Missing another timeout

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
+  assert.strictEqual(this.player.ads.state, 'content-playback');
+});
+
+QUnit.test('adskip in postroll? transitions to content-playback and fires ended', function(assert) {
+  assert.expect(8);
+  assert.strictEqual(this.player.ads.state, 'content-set');
+  this.player.trigger('adsready');
+  assert.strictEqual(this.player.ads.state, 'ads-ready');
+  this.player.tech.trigger('play');
+  this.player.trigger('adtimeout');
```
dmlap: Missing timeout

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
+  this.player.trigger('ended');
+  assert.strictEqual(this.player.ads.state, 'postroll?');
+  this.player.ads.startLinearAdMode();
+  this.player.ads.snapshot.ended = true;
+  this.player.ads.endLinearAdMode();
+  assert.strictEqual(this.player.ads.state, 'content-resuming');
+  assert.strictEqual(spy.callCount, 0, 'we should not have gotten an ended event yet');
+  this.clock.tick(2000);
```
dmlap: Note to self: this seems long, double check.

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
-  if (!/phantomjs/i.test(window.navigator.userAgent)) {
-    cbs.pop().call(window);
-    equal(ended, 1, 'we should have fired ended from the timeout cbs');
-  }
-
-  window.setTimeout = oldtimeout;
-  window.clearTimeout = oldclear;
-});
-
-test('an ended event is not fired in content-resuming via a timeout if fired naturally', function() {
```
dmlap: Looks like this test got dropped.

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,732 @@
...
+  this.player.ads.skipLinearAdMode();
+  assert.strictEqual(this.player.ads.state, 'ad-playback');
+  assert.strictEqual(spy.callCount, 0, 'adskip event should not trigger when skipLinearAdMode called in ad-playback state');
+  this.player.ads.endLinearAdMode();
+  assert.strictEqual(this.player.ads.state, 'content-resuming');
+  assert.strictEqual(this.player.ads.triggerevent, 'adend', 'The reason for content-resuming should have been adend');
+  this.player.trigger('playing');
+  assert.strictEqual(this.player.ads.state, 'content-playback');
+  assert.strictEqual(this.contentPlaybackSpy.callCount, 1, 'A content-playback event should have triggered');
+  assert.strictEqual(this.contentPlaybackReason(), 'playing', 'The reason for content-playback should have been playing');
```
dmlap: It's kind of hard to keep track of the flow in these last couple tests without a bit of whitespace.

package.json
misteroneill: I thought it was `~`.

I've run into a lot of issues in the past with `^` bringing in breaking changes (despite semver saying minor versions should be backward-compatible); sometimes multiple dependency layers deep.

That said, I'm fine with changing these. It ought to be a company/project-wide standard, though.

src/videojs.ads.js
misteroneill: I would think so, yes.

test/videojs.ads.snapshot.js
misteroneill: I don't recall. I'll verify that it is and investigate.

test/videojs.ads.snapshot.js
misteroneill: I believe I was getting errors triggering events on the player. I'll see if it's still necessary.

test/videojs.ads.snapshot.js
misteroneill: Good call!

test/videojs.ads.test.js
misteroneill: Hmmm, good question. I'll bring it back.

test/videojs.ads.test.js
misteroneill: Nope, good catch.

test/videojs.ads.test.js
misteroneill: Yeah, I agree. I'll go back and add some in reasonable places.

test/videojs.ads.test.js
```diff
@@ -1,1025 +1,745 @@
...
+  assert.strictEqual(spy.callCount, 1, 'a preroll triggers adstart');
+});
+
+QUnit.test('player has the .vjs-has-started class once a preroll begins', function(assert) {
+  assert.expect(1);
+  this.player.trigger('adsready');
+
+  // This test fails unless the tech is directly triggered here. Why? Bug in
+  // video.js 5.0?
+  this.player.trigger('play');
```
misteroneill: As the comment says, this test fails because this `trigger('play')` call doesn't seem to do what is expected. `this.player.tech_.trigger('play')` does work, however.

misteroneill: `.play()` doesn't work either; though, it does in video.js proper.

dmlap:
This was really tough to review. I know the timeout stuff was a pain to convert but it would have been a lot easier if we could have split the upgrade + refactor.

test/videojs.ads.test.js
dmlap: Oh, I missed that comment. It does seem like we need to track that down since `tech` is made "private" in a later RC.

src/videojs.ads.js
dmlap: Ok. Good find if it was an existing issue, just wanted to make sure it wasn't a subtle version incompatibility.

test/videojs.ads.test.js
dmlap: I think this could be reduced to [1000](https://github.com/videojs/videojs-contrib-ads/blob/master/src/videojs.ads.js#L641). It's fine as-is, though.

misteroneill:
Yeah, that's fair.

There were so many test failures related to subtle differences in things that are now async that were not previously - as well as the order and structure of events between video.js and contrib-ads - that it would've been a big thing regardless.

package.json
gkatsev: With sinon, I would even just get rid of the prefix altogether. They broke too many times on minor and patch releases to trust them.
lodash I think is fine to use `^`.

test/videojs.ads.test.js
misteroneill: I'll change this and other timeouts. It makes sense to me that they should only be ticking enough to trigger the behavior - otherwise, it might hide weird edge-cases or something.

package.json
misteroneill: Interesting. I've never had a problem with Sinon at all.

test/videojs.ads.snapshot.js
misteroneill: I'm adding a comment, but it's necessary because the "canplay" event causes the `restorePlayerSnapshot` to be called which queues up 20 timeouts at 50ms each (hence the 1000ms).

test/videojs.ads.snapshot.js
misteroneill: Looked again and it's 1250: 20 `tryToResume` timeouts at 50ms each + `resumeEndedTimeout` at 250ms.

test/videojs.ads.test.js
misteroneill: Verified that 1000 is fine here.

test/videojs.ads.snapshot.js
misteroneill: Going to note here that this is because triggering "play" on the player does not cause anything to happen. Rather, these events bubble up from the tech.

test/videojs.ads.test.js
misteroneill: I'm not sure, but it exists because we need to wait for the `player.ads.cancelPlayTimeout` (which is a `defer`, i.e. 1ms) to resolve.

test/videojs.ads.test.js
misteroneill: I don't know that it _was_ required in actuality.

test/videojs.ads.test.js
misteroneill: Yeah, looks like I thought it was duplicated since it's so similar to the previous one. Bringing it back!

misteroneill:
This is good for review once again.

nickygerritsen:
I'm using this PR for implementing ads in 5.0, but I have a suggestion:

Could you add a `main` section to `package.json`. Then I can use a nice `import 'videojs-contrib-ads'` instead of `import '../node_modules/videojs-contrib-ads/dist/videojs.ads.js'`.

Maybe this only makes sense after releasing?

dmlap:
Would be nice to get @nickygerritsen's suggestions in here too but LGTM either way.

dist/videojs.ads.global.js
```diff
@@ -0,0 +1,813 @@
...
+      }
+
+      player.play();
+    });
+
+    // replace the ad initializer with the ad namespace
+    player.ads = {
+      state: 'content-set',
+
+      // Call this when an ad response has been recieved and there are
```
nickygerritsen: Although not changed in this PR, recieved does not seem as a valid work to me ;). This is already wrong in the `src` file of course.

The same typo is there 15 lines lower.

misteroneill: You're right - "i" before "e" except after "c"! :)

misteroneill:
Closing in favor of #121

PR #131, opened 2015-11-20
VJS5 - resuming video tech fix

alex-phillips:
NOTE: This is for the `vjs-5-upgrade` branch only!!

Because the tech may have changed from the original video to the ad video, we need to make sure we have the correct `tech` object before attempting to resume.

Ex: The video playing is an RTMP stream using the Flash player, the video ad playing was an MP4 (so HTML5), when attempting to resume, we were setting the `tech` variable BEFORE switching the source back to the RTMP stream so the video was never resuming from where it left off.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -246,6 +246,9 @@ var
       // determine if the video element has loaded enough of the snapshot source
       // to be ready to apply the rest of the state
       tryToResume = function() {
+        // Tech may have changed depending on the differences in sources of the
+        // original video and that of the ad
+        tech = player.el().querySelector('.vjs-tech');
         if (tech.readyState > 1) {
           // some browsers and media aren't "seekable".
           // readyState greater than 1 allows for seeking without exceptions


misteroneill:
Thanks for the PR! This seems reasonable to me, but it seems like a potentially obscure issue and would really benefit from some test(s) when you have some time.

Also, just a word of warning, the `vjs-5-upgrade` branch will probably get merged into `master` in the near future; so, you'll need to rebase this at that point.

alex-phillips:
@misteroneill I didn't even think about opening the PR to the `vjs-5-upgrade` branch. If I close this one and reopen under that branch, could it get merged in before the upgrade path is merged into master?

I'll look into adding a test case as well. Thanks for the help!

gkatsev:
We could definitely try and get it merged if you open it against that. Worst case, you'll just need to just point PR at master instead (if we end up merging before you're done)

misteroneill:
@alex-phillips PR #121 has been merged, so feel free to rebase your branch! Once a test case is in, we can look at merging this. Thanks!

alex-phillips:
Thanks guys! In order for a test to work properly, I need a source RTMP stream for the example code. I see you're using a test `mp4` from w3.org. Any thoughts or preferences on where to reference an RTMP stream?

alex-phillips:
So I may be doing something wrong, but I thought I'd bring it up in case someone has some insight into the issue, but inside of the test suite, if I pass in an RTMP source (valid, it works in manual tests as well as live on our site), the tests kick back and says `VIDEOJS: ERROR: (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) No compatible source was found for this video.`.

Is flash media able to be run/tested inside of the test suite?

misteroneill:
@alex-phillips I assume you mean on Travis. The answer is "no" unfortunately because the tests currently run in PhantomJS.

A lot of the tests for snapshots do some pretty heavy-handed mocking; so, you might just create your own instance of the `Html5` tech and replace the existing one with it. Even if that means just dumping a new DOM element in the player!

If this turns out to be more trouble than it's worth just say so and we are probably fine to merge this - the code shouldn't be caching DOM references that may be accessed within async processes anyway.

Thanks!

gkatsev:
LGTM

misteroneill:
Merging this because the existing code was relying on cached DOM references in an asynchronous callback, which is... risky, to say the least. We're confident it won't break anything.

misteroneill:
Thanks @alex-phillips!

Issue #132, opened 2015-12-01
Issue with currentSrc() and MSE and SourceHandlers


gkatsev:
In Videojs 5.2.2 we made it so that source handlers will return the saved source rather than the src from the video element. This is important for projects like contrib-hls, so that they can return a proper m3u8 url instead of a funky blob url.
This causes issues with ads because the snapshot saves the `currentSrc()` only but when restoring, it first checks against `player.src()` which is still returning the `blob` url. Since those are different, it tries to restore the video even though it shouldn't do that.
Adblockers apparently can sometimes interfere with `currentSrc()` getting set properly for native playback, which means that you can have a valid usecase for having a different `src()` and `currentSrc()` and needing to restore the snapshot. See https://github.com/videojs/videojs-contrib-ads/blob/master/test/videojs.ads.snapshot.js#L343-L375

I think the solution here would be to save both the `currentSrc()` *and* `src()` values and compare both of them before restoring the snapshot.
In the meantime, once https://github.com/videojs/video.js/issues/2852 is in, before starting ads, the integration (or contrib-ads itself, even), could set the `src('')` to reset the player so the source will always be different and we'd always be restoring it.

misteroneill:
Post video.js 5.2.2, this is causing an issue where the video's poster is displayed after an ad ends and content resumes because it detects that the `src` has changed and resets the player.

misteroneill:
Closed by #133

PR #133, opened 2015-12-02
Update to 5.3.x / fix srcChanged logic.

misteroneill:
This updates us to video.js 5.3.x as well as fixing (I think...) the logic around restoring video source after an ad plays.

Related to #132


gkatsev:
We should make sure that the changes to `srcChanged` also fix #132. It'll be silly to be fixing it multiple times.

misteroneill:
Agreed :100:%

misteroneill:
I'm confident that the three cases here are sufficiently tested by existing tests; so, I think this is ready for review.

PR #134, opened 2015-12-02
Be More Dry

misteroneill:
This dries up a lot of code that was old and no longer needed. It builds on #133.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -9,22 +9,6 @@
 var
 
   /**
-   * Copies properties from one or more objects onto an original.
-   */
-  extend = function(obj /*, arg1, arg2, ... */) {
-    var arg, i, k;
-    for (i=1; i<arguments.length; i++) {
-      arg = arguments[i];
-      for (k in arg) {
-        if (arg.hasOwnProperty(k)) {
-          obj[k] = arg[k];
-        }
-      }
-    }
-    return obj;
-  },
-
-  /**
    * Add a handler for multiple listeners to an object that supports addEventListener() or on().
    *
    * @param {object} obj The object to which the handler will be assigned.
@@ -364,10 +348,7 @@ var
   adFramework = function(options) {
     var
       player = this,
-
-      // merge options and defaults
-      settings = extend({}, defaults, options || {}),
-
+      settings = videojs.mergeOptions(defaults, options),
       fsmHandler;
 
     // prefix all video element events during ad playback
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -3,30 +3,12 @@
  *
  * Common code to support ad integrations.
  */
-(function(window, document, videojs, undefined) {
-"use strict";
+(function(window, videojs, undefined) {
+'use strict';
 
 var
 
   VIDEO_EVENTS = videojs.getComponent('Html5').Events,
-
-  /**
-   * Defer an action until the call stack is clear.
-   *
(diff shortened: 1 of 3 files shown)


src/videojs.ads.js
```diff
@@ -382,75 +270,68 @@ var
...
+        event.cancelBubble = true;
+        event.isPropagationStopped = returnTrue;
+        player.trigger({
+          type: type + event.type,
+          state: player.ads.state,
+          originalEvent: event
+        });
+      };
+
+      player.on(videoEvents, function(event) {
```
gkatsev: Can you still name this function `redispatch`?

src/videojs.ads.js
```diff
@@ -373,75 +270,68 @@ var
...
+        event.cancelBubble = true;
+        event.isPropagationStopped = returnTrue;
+        player.trigger({
+          type: type + event.type,
+          state: player.ads.state,
+          originalEvent: event
+        });
+      };
+
+      player.on(videoEvents, function redispatch(event) {
```
gkatsev: I assume the redispatch function is the same just moved? The diff makes it a bit hard to see it fully/exactly.

src/videojs.ads.js
```diff
@@ -479,249 +369,248 @@ var
...
-                this.state = 'preroll?';
-                cancelContentPlay(player);
-              },
-              'adskip': function() {
-                this.state = 'content-playback';
-              },
-              'adserror': function() {
-                this.state = 'content-playback';
-              }
+      var fsm = {
```
gkatsev: and here as well just indentation/whitespace change?

src/videojs.ads.js
```diff
@@ -3,98 +3,12 @@
...
-    switch (type) {
-      case '[object String]':
-        register(obj, events, handler);
-        break;
-      case '[object Array]':
-        for (i = 0, ii = events.length; i<ii; i++) {
-          register(obj, events[i], handler);
-        }
-        break;
-      case '[object Object]':
```
gkatsev: were we ever even using this case in the code? This would be the only thing that possibly could break.

src/videojs.ads.js
```diff
@@ -359,13 +260,9 @@ var
...
   adFramework = function(options) {
-    var
-      player = this,
-
-      // merge options and defaults
-      settings = extend({}, defaults, options || {}),
-
-      fsmHandler;
+    var player = this;
+    var settings = videojs.mergeOptions(defaults, options);
```
gkatsev: do we need the `|| {}` or does `mergeOptions` do that for us?

src/videojs.ads.js
misteroneill: No, I checked - only strings and arrays!

src/videojs.ads.js
misteroneill: Yeah, it got moved to be an inline callback and the construction of the `videoEvents` was greatly simplified.

src/videojs.ads.js
misteroneill: Yeah.

gkatsev:
Adding `w=1` to the query params of the URL made this PR a whole lot more readable.

misteroneill:
Yeah, sorry about that. :smile:

src/videojs.ads.js
misteroneill: Don't need it.

gkatsev:
LGTM

PR #135, opened 2015-12-09
Update .npmignore to use generator version

misteroneill:
This avoids things like editor files getting included in the package.


misteroneill:
Just gonna merge this since it should be harmless (beneficial even!) and get a new tag out.

PR #136, opened 2015-12-10
Race canplay event with a setTimeout.

gkatsev:
In some browsers, like firefox with MSE, canplay doesn't fire at the
correct time. So, we need to race it against a setTimeout so that we
would actually try to resume if canplay doesn't fire correctly.
In the tryToResume function we then clear out the event handler and the
timeout depending on which came back first.

Also, see https://github.com/videojs/videojs-contrib-hls/issues/469

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -148,9 +148,22 @@ var
       // determine if the video element has loaded enough of the snapshot source
       // to be ready to apply the rest of the state
       tryToResume = function() {
+
+        // tryToResume can either have been called through the `contentcanplay`
+        // event or fired through setTimeout.
+        // When tryToResume is called, we should make sure to clear out the other
+        // way it could've been called by removing the listener and clearing out
+        // the timeout.
+        player.off('contentcanplay', tryToResume);
+        if (player.ads.tryToResumeTimeout) {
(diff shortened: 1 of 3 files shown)


src/videojs.ads.js
```diff
@@ -211,7 +224,10 @@ var
       // safari requires a call to `load` to pick up a changed source
       player.load();
       // and then resume from the snapshots time once the original src has loaded
+      // in some browsers (firefox) `canplay` may not fire correctly.
+      // Reace the `canplay` event with a timeout.
       player.one('contentcanplay', tryToResume);
+      player.ads.tryToResumeTimeout = player.setTimeout(tryToResume, 2000);
```
gkatsev: I used 2 seconds here. I'm not sure what this should actually be.

dmlap: Probably should suffix this with an underscore so it's obvious it's an internal variable.

dmlap:
One comment about variable naming but otherwise LGTM

src/videojs.ads.js
gkatsev: Makes sense.

gkatsev: Fixed.

PR #137, opened 2015-12-14
after postoll, seek to end

bcvio:
Sometimes `snapshot.currentTime` is just  slightly bigger than the actual reported duration, and restoring to that raw value will cause the player to hang in a loading state.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -61,6 +61,7 @@ var
         currentSrc: player.currentSrc(),
         src: player.src(),
         currentTime: player.currentTime(),
+        duration: player.duration(),
         type: player.currentType()
       };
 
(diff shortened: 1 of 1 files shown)


src/videojs.ads.js
```diff
@@ -113,12 +114,14 @@ var
           updateEnded = function() {
             ended = true;
           };
-        player.currentTime(snapshot.currentTime);
```
dmlap: So `snapshot.currentTime` was not returning the actual video duration in some cases before a postroll? Any details on where/why this was occurring?

misteroneill: This seems really strange to me. It may address the issue, but it doesn't feel like the right path. It seems counter-intuitive that `currentTime` would be different after an ad. Could this be exposing an issue in HLS/MSE? Has this been tried with a plain old MP4?

bcvio: This is in fact caused by an issue with duration reporting in HLS/MSE. In this case, the duration() is returning slightly different inconsistent values throughout the content playback.

dmlap: Reporting duration can be trickier with MSE or segmented file formats since accurate duration information is only available once a good portion of the video has been buffered. That has been the case anytime we play HLS though and this is the first time I've heard of this. Is it a bug in the new version of contrib-hls, a bug/fact-of-life with MSE, or reasonable behavior that contrib-ads needs to account for?

misteroneill: If this is a fact of life with MSE, I wonder if there ought to be some detection of the type of source in the player or an explicit option that flags the player as using a segmented format.

misteroneill:
I believe this issue is no longer relevant due to fixes that were added to `videojs-contrib-hls`. If that's wrong, please re-open!

PR #138, opened 2015-12-15
Fix example page

gkatsev:


sprice:
Just ran into this issue myself. This is the exact same fix I came up with.

imbcmdth:
LGTM :pager:

PR #141, opened 2016-01-20
startLinearAdMode only triggers adstart from appropriate states

incompl:


gkatsev:
I wonder if we should add a test for this.
Otherwise, LGTM.

incompl:
@gkatsev Added a test, would like review on that too, not sure it's the most useful test or how else we might test it.

misteroneill:
Just a thought, but I wonder if this ought to be a whitelist instead of a blacklist?

gkatsev:
Isn't this a whitelist? You can only `adstart` from these 3 states?

misteroneill:
Haha, you're right, I read it too fast. Ignore me. Carry on. :laughing:

gkatsev:
LGTM

incompl:
@gkatsev @misteroneill Looks like this is good to go, merge away!

PR #142, opened 2016-01-25
Backport fix for MSE HLS src/currentSrc

forbesjo:
This PR is to backport #133 to fix using video.js 5.0 with MSE HLS and the 4.0 to 5.0 compatibility script.


misteroneill:
LGTM :+1:

PR #143, opened 2016-01-27
Don't restore player snapshot after ad is done in adonly mode

incompl:


misteroneill:
A thought: I wonder if we ought to store some kind of state on the `player.ads` object that contrib-ads creates rather than querying the DOM.

incompl:
@misteroneill I thought about that, but this felt most consistent with what I'm seeing in this plugin and in other plugins such as analytics. Not sure using the player classlist as the universal state tracker is the best architectural choice but it seems to be used consistently.

incompl:
Realized this shouldn't go in contrib-ads and came up with a different approach. Closing this PR.

PR #146, opened 2016-02-25
remove class vjs-ad-loading in ad-ready? state

marguinbc:
While we are waiting for the ima3 sdk to load in ad-ready? we aren't waiting for an actual ad to load as vjs-ad-loading implies


marguinbc:
Realized a better way to achieve the end result I was looking for - probably best to leave this as is so as not to introduce and unwanted side effects.

PR #147, opened 2016-03-03
Allow ad impl to skip wait for preroll if it is known there is no preroll coming

incompl:


misteroneill:
I wonder about the naming, but perhaps more information on the purpose of this would be beneficial?

incompl:
@misteroneill You made me realize this should be in the README. I added it. What do you think?

misteroneill:
Sounds good to me. Would be good to get another set of eyes on this, though.

gkatsev:
LGTM

PR #148, opened 2016-03-21
Live ads support

incompl:
Depends on #147

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -68,8 +68,7 @@ var
     if (videojs.browser.IS_IOS && isLive(player)) {
       // Record how far behind live we are
       currentTime = player.currentTime() - player.seekable().end(0);
-    }
-    else {
+    } else {
       currentTime = player.currentTime();
     }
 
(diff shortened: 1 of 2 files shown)


misteroneill:
LGTM 👍

src/videojs.ads.js
```diff
@@ -43,13 +43,35 @@ var
   },
 
   /**
+   * Returns a boolean indicating if given player is in live mode.
+   */
+  isLive = function(player) {
+    if (player.duration() === Infinity) {
+      return true;
+    } else if (videojs.browser.IOS_VERSION === "8" && player.duration() === 0) {
```
dmlap: @gkatsev Should video.js be handling this discrepancy? Seems like the kind of problem we try to abstract over.

gkatsev: Sounds like iOS is once again doing stuff that is outside of the spec (https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-duration) however, yes, that's totally something that videojs itself should be handling.

incompl: I could open an issue against VJS for this and link to it in a comment. Sound like a plan?

gkatsev: That works. Or, if you want to make a PR against stable, we can get a patch release out chikchak.

incompl: Opened issue and added comment

PR #150, opened 2016-03-29
Rollback nopostroll logic in postroll?

incompl:


dmlap:
It looks like there is still documentation for the `nopostroll` event and a listener elsewhere in this file.

test/videojs.ads.test.js
```diff
@@ -1005,13 +1005,13 @@ QUnit.test('ad impl can notify contrib-ads there is no preroll', function(assert
 
 });
 
-QUnit.test('ad impl can notify contrib-ads there is no postroll', function(assert) {
+// QUnit.test('ad impl can notify contrib-ads there is no postroll', function(assert) {
```
dmlap: For the future, `QUnit.skip` achieves this and leaves a reminder around to re-enable the test later.

PR #151, opened 2016-03-30
Add tests for ended event, in light of recent bug when implementing nopostroll

incompl:


test/videojs.ads.test.js
```diff
@@ -1005,13 +1005,45 @@ QUnit.test('ad impl can notify contrib-ads there is no preroll', function(assert
...
+});
+
+QUnit.test('ended event is sent with postroll', function(assert) {
+
+  var ended = sinon.spy();
+
+  this.player.tech_.el_.ended = true;
+  this.player.on('ended', ended);
+  this.player.ads.state = 'content-playback';
+  this.player.trigger('contentended');
```
gkatsev: why `contentended` and not `ended`?

incompl: The test is making sure that ended is sent. If we send it manually in the test, I think that would invalidate the test.

gkatsev: Maybe we want another test that does trigger `ended` manually since it could still be useful to make sure that the redispatch is working correctly. Since, depending on what state we're in, `ended` should end up changing to be `adended` or `contentended`.

incompl: I agree it would be good to test redispatch more broadly. For this PR, my goal is just to make a regression test for the bug that was introduced yesterday.

PR #157, opened 2016-04-20
document disableNextSnapshotRestore option

bcvio:


README.md
```diff
@@ -229,6 +229,17 @@ player.ads.contentSrc = 'movie-high.mp4';
 player.src('movie-high.mp4');
 ```
 
+### disableNextSnapshotRestore
+Prevents the next snapshot restoration.
+
+If you need to change the video source during ad playback, you can use _disableNextSnapshotRestore_ to prevent videojs-contrib-ads to restore to the previous video source.
```
incompl: A little clearer as "to prevent videojs-contrib-ads from restoring the previous video source"

incompl:
LGTM

PR #158, opened 2016-04-26
Fix prefixing events when player is preload none

incompl:
Related: https://github.com/videojs/video.js/pull/2957 (necessary for this to work in HLS after the src has been changed by a third party)

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -311,7 +311,10 @@ var
     // when truthy, instructs the plugin to output additional information about
     // plugin state to the video.js log. On most devices, the video.js log is
     // the same as the developer console.
-    debug: false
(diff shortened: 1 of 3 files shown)


src/videojs.ads.js
```diff
@@ -430,9 +429,28 @@ var
         if (player.ads.state !== 'ad-playback') {
           player.trigger('adskip');
         }
+      },
+
+      stitchedAds: function(arg) {
+        if (arg) {
```
misteroneill: Should this be `arg !== undefined`? It doesn't look like there would be a way to set it to `false`.

incompl: Fixed

misteroneill:
LGTM

PR #163, opened 2016-05-05
Check seekable().length before calling seekable().end(0)

incompl:


src/videojs.ads.js
```diff
@@ -68,7 +68,12 @@ var
 
     if (videojs.browser.IS_IOS && isLive(player)) {
       // Record how far behind live we are
-      currentTime = player.currentTime() - player.seekable().end(0);
+      try {
+        currentTime = player.currentTime() - player.seekable().end(0);
```
gkatsev: Why not
```js
if (player.seekable().length > 0) {...}
```
Error out on `seekable()` in generaly?

incompl: I changed it to that

dmlap:
LGTM

misteroneill:
LGTM

PR #165, opened 2016-05-18
After postroll, seek to the actual player duration. If we are beyond …

bcvio:
…snapshot.ended and the `postroll?` state has already been consumed, say good-bye

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -738,7 +738,7 @@ var
               }
             },
             'contentended': function() {
-              if(player.ads.snapshot && player.ads.snapshot.ended){
+              if (player.ads.snapshot && player.ads.snapshot.ended) {
                 // player has already been here. content has really ended. good-bye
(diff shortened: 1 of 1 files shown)


incompl:
LGTM

This would be a good thing to test with the [player-tester](https://bithub.brightcove.com/videocloud/player-tester)!

src/videojs.ads.js
```diff
@@ -738,6 +738,10 @@ var
               }
             },
             'contentended': function() {
+              if(player.ads.snapshot && player.ads.snapshot.ended){
```
gkatsev: spaces after `if` and before `{`.

misteroneill: I wonder also if it'd make more sense to do:

```
if (!player.ads.snapshot || !player.ads.snapshot.ended) {
  this.state = 'postroll?';
}
```

Just a thought.

PR #167, opened 2016-05-25
fix for double ad-end event

bcvio:


src/videojs.ads.js
```diff
@@ -617,19 +617,16 @@ var
             if (!videojs.browser.IS_IOS && player.duration() === Infinity) {
               player.volume(this.preAdVolume_);
             }
-
-            // trigger 'adend' as a consistent notification
-            // event that we're exiting ad-playback.
-            if (player.ads.triggerevent !== 'adend') {
-              player.trigger('adend');
```
incompl: What effect does removing this have?

bcvio: In a flash environment, where we rely on ExternalInterface communication, when leaving `ad-playback`, this would trigger a second `adend` event. 
For a consistent notification of exiting `ad-playback` state, it is really enough to trigger the `adend` in the eventuality of `adserror` .

incompl: So, because 'adend' was triggered both in endLinearAdMode and in ad-playback.leave, that was why there were duplicate ad-ends? And you're removing this one and making an error-specific one. Just want to make sure I understand what's going on here.

bcvio: Yes, that is why we sometimes have double `adend` events.
We just want to make sure we notify with an `adend` every time we leave `ad-playback`. The only other time that happens is when `adserror` is triggered, thus making sense to move trigger('adend')

incompl: Talked to @bcvio out of band and the question is answered.

PR #174, opened 2016-06-08
Fix android timing issues

incompl:
This fixes a couple issues on Android where an ad would not pause the video and/or the content would not resume correctly after the ad.


misteroneill:
LGTM

apesant:
Hi,

I just noticed that this fix seems to break the getPlayerSnapshot method on iOS. There is an exception being thrown. Could you try it?

incompl:
Thanks for reporting this, @apesant. I've pushed a fix for that issue to this pull request.

apesant:
Thanks, it seems to work now :)
Do you have an ETA about when this will be merged and released? I don't need an exact date, a rough approximation is fine!

incompl:
I'm guessing within the next week or two.

pagarwal123:
Looks good to me now

PR #179, opened 2016-07-20
ios content flash

incompl:


ldayananda:
This seems like a nice, modular fix. Still very likely to cause issues once we test on a lot of devices.

Was hiding the player required specifically for the iOS fullscreen mode case?

src/videojs.ads.js
```diff
@@ -25,6 +25,28 @@ var
       // another cancellation is already in flight, so do nothing
       return;
     }
+
+    // Avoid content flash on iOS
+    if (videojs.browser.IS_IOS) {
+
+      // A placeholder black box will be shown in the document while the player is hidden.
+      var placeholder = document.createElement('div');
+      placeholder.style.width = player.width() + 'px';
```
gkatsev: you probably want to use `player.currentWidth()` and `player.currentHeight()` here with the caveat that it will mean a new minimum supported version of videojs (5.9).

incompl: Would that make it work better?

gkatsev: Yes, `player.width` and `player.height` are both the *configured* values for the size of the player. They're match `videoEl.getAttribute('width')` which means if you don't have a size configured, you'll end up getting `0px` here.

incompl: Did both with fallback

src/videojs.ads.js
```diff
@@ -29,10 +29,13 @@ var
     // Avoid content flash on iOS
     if (videojs.browser.IS_IOS) {
 
+      var width = player.currentWidth ? player.currentWidth() : player.width();
```
gkatsev: 👍

src/videojs.ads.js
```diff
@@ -25,6 +25,32 @@ var
...
+      placeholder.style.height = height + 'px';
+      placeholder.style.background = 'black';
+      player.el_.parentNode.insertBefore(placeholder, player.el_);
+
+      // Hide the player. While in full-screen video playback mode on iOS, this
+      // makes the player show a black screen instead of content flash.
+      player.el_.style.display = 'none';
+
+      // Unhide the player and remove the placeholder once we're ready to move on.
+      player.one(['adplaying', 'adtimeout', 'adserror', 'adscanceled', 'adskip',
```
bcvio: Is `adplaying` an event?

incompl: Yup! It may be only when the ad plays in the content element, but that's exclusively when this code runs.

bcvio:
LGTM

incompl:
Team agreed to split [this remaining playlist foc](https://jira.brightcove.com/browse/BC-37036) into a separate issue and merge this improvement.

draja91:
LGTM

PR #180, opened 2016-07-21
Prevent replay after postroll

incompl:


src/videojs.ads.js
```diff
@@ -147,33 +143,19 @@ var
           player.currentTime(snapshot.ended ? player.duration() : snapshot.currentTime);
         }
 
+        // Android reports duration Infinity right after switching source, so we need
+        // this to determine if ad was a postroll.
+        var androidPostroll = snapshot.ended &&
```
dmlap: Can we add more info to the snapshot to identify this? This condition seems racy to me.

src/videojs.ads.js
```diff
@@ -455,7 +437,7 @@ var
       // Returns a boolean indicating if given player is in live mode.
       // Can be replaced when this is fixed: https://github.com/videojs/video.js/issues/3262
       isLive: function(player) {
-        if (player.duration() === Infinity) {
+        if (player.currentTime() > 0 && player.duration() === Infinity) {
```
dmlap: While it might only be wrong in a very small number of cases, this additional condition seems dangerous to me. I worry we'd get some very hard to reproduce bug reports.

src/videojs.ads.js
incompl: Big picture, I want contrib-ads to have a clear idea of what's going on, rather than try to infer the state of things everywhere. This is one of the advantages of the architecture idea where preroll / midroll / posteroll are inherited methods: there would be no confusion about where we're at. As for a short term / incremental improvement, I'm not sure what to suggest.

src/videojs.ads.js
incompl: This matches what we're doing here, do you have the same concern in that case? https://github.com/videojs/video.js/pull/3083

dmlap:
Beauty. LGTM.

For posterity: @incompl wasn't able to repro the `ended` issue on iOS 8.1 anymore so we removed the weird workaround.

apadhye:
QA LGTM

PR #181, opened 2016-07-25
Update README.md

timonbandit:
Your example is contain an error;
on-ads-end event should be "adend" instead of "adended".
Thanks)


gkatsev:
Thanks for the PR, however, this example is correct. `adended` is triggered when an ad finishes playing (in some cases, like when the video element is being re-used for ads like in the example), then we want to call `endLinearAdMode` which triggers `adend`.

PR #182, opened 2016-07-26
Fix no ended event after content source is changed.

ldayananda:
# Description
Previously, if a new source is given to the player after the 'ended' event has been triggered, the 'ended' event will not fire again.

# Changes Proposed
Removing reference to snapshot when a contentupdate is observed during content playback.

### DIFF
--- a/src/videojs.ads.js
+++ b/src/videojs.ads.js
@@ -757,6 +757,12 @@ var
               } else {
                 this.state = 'ads-ready?';
               }
+              // When a new source is loaded into the player, we should remove the snapshot
+              // to avoid confusing player state with the new content's state
+              // i.e When new content is set, the player should fire the ended event
+              if (this.snapshot && this.snapshot.ended) {
+                this.snapshot = null;
+              }
             },
             'contentended': function() {
               if (player.ads.snapshot && player.ads.snapshot.ended) {


src/videojs.ads.js
```diff
@@ -757,6 +757,12 @@ var
               } else {
                 this.state = 'ads-ready?';
               }
+              // When a new source is loaded into the player, we should remove the snapshot
+              // to avoid confusing player state with the new content's state
+              // i.e When new content is set, the player should fire the ended event
+              if (this.snapshot && this.snapshot.ended) {
```
incompl: Looks good. Any reason to keep the snapshot if it's not ended?

ldayananda: Essentially damage control. I saw differences in the event flow when I removed the snapshot regardless of whether it is ended or not.

incompl: Oh geez. Well, guess all we can do is test it.

incompl:
LGTM

apadhye:
QA LGTM

StaymanHou:
We ran into a problem, which led me here.
Our player is using videojs 5, videojs-contrib-ads, and modified version of [googleads-ima-flash
](https://github.com/googleads/googleads-ima-flash). We noticed that in certain conditions, the ended event is not triggered.

Step to reproduce:
1) Set the tag to a IMA tag which returns no ads or returns wrappers that has no fill
2) Start the playback
3) The first `ended` event will fire. After that replace the video content src. This is probably optional. In our use case, the player will continue to play the next video in a playlist
4) Reset the tag to another one which has no ads, or keep the previous one
5) Start the playback
6) This time it doesn't trigger the `ended` event

My workaround:
Reverting the change in commit aa46ca3d06b44d99e44bcff60d563961845d2e69 seems to resolve the issue.

ldayananda:
Thank you for your comment @StaymanHou.

I have not been successful in reproducing the issue you have described. If you could please open a new issue with a reduced example, it would help us investigate the issue further.

PR #187, opened 2016-09-21
Don't send playing before prerolls, just after (and major commenting for redispatch)

incompl:


src/videojs.ads.js
```diff
@@ -337,43 +338,119 @@ var
...
+            player.ads.videoElementRecycled()) {
+
+          // Triggering an event prevents the unprefixed one from firing.
+          // "adcontentplaying" is only seen in this very specific condition.
+          triggerEvent('adcontent', event);
+          return;
+
+        // When ad is playing in content tech, prefix everything with "ad".
+        // This block catches many events such as emptied, play, timeupdate, and ended.
+        } else if (player.ads.state === 'ad-playback') {
```
ldayananda: I like this change and think this makes the event flow much easier to flow. However, this is likely to affect other plugins relying on these events and so should be thoroughly documented.

incompl: Yeah I already know of one place where the change to "playing" would break something else. This will be tricky

ldayananda:
LGTM

pagarwal123:
QA: LGTM

PR #193, opened 2016-10-04
Break out features into separate files

incompl:


src/redispatch.js
```diff
@@ -0,0 +1,141 @@
...
+expect despite the presense of ads. For example, an integrator would expect
+an `ended` event to happen once the content is ended. If an `ended` event is sent
+as a result of an ad ending, that is a bug. The `redispatch` method should recognize
+such `ended` events and prefix them so they are sent as `adended`, and so on with
+all other player events.
+*/
+
+import videojs from 'video.js';
+
+let player;
```
gkatsev: this would get shared between all players.

gkatsev: Looking at the code, it *might* be ok as is but we should probably try to get rid of this if we can.

src/videojs.ads.css
```diff
@@ -1,16 +0,0 @@
-/**
- * videojs.ads.css
```
gkatsev: why is this file removed?

incompl: Moved to plugin.scss

src/redispatch.js
incompl: I have been thinking about the best way to do this. One option is to make this a class and you pass a player to the constructor. What do you think?

src/videojs.ads.css
gkatsev: Oh, I see. Missed it from the other PR.

src/redispatch.js
gkatsev: Is the only reason we're using `player` here for `prefixEvent`? What about passing `player` through to it directly.

incompl: It ends up being a lot of boilerplate that I want to avoid. Of course, a class is a lot of boilerplate too.

gkatsev: Maybe wrap it all in a function?

incompl: Just ended up passing player directly, at least it's explicit and clear

src/cancelContentPlay.js
```diff
@@ -0,0 +1,57 @@
+/*
+This feature makes sure the player is paused during ad loading.
+
+It does this by pausing the player immediately after a "play" where ads will be requested,
+then playing after the ad is done if we paused the player in this way.
```
ldayananda: This technically doesn't unpause the player in this file

bcvio:
LGTM

PR #195, opened 2016-10-06
Fix for content-play behind ad on Android

vdeshpande:
Proposal:
This is a proposal fix for preventing content from playing in the background while an Ad is playing on an Android device within a playlist.

The current issue is that if you have an auto-advance playlist containing HLS videos, the duration of the video when the Ad starts is given as Infinity erroneously on Android devices(I have tested this on v5.x and v6.x), and the content does not stop when a pre-roll is playing.

The idea is to pause playback of the content before we signal the ad plugin to play a pre-roll.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -266,6 +266,11 @@ const contribAdsPlugin = function(options) {
           player.ads.adTimeoutTimeout = window.setTimeout(function() {
             player.trigger('adtimeout');
           }, settings.prerollTimeout);
+          // We will try to pause content here. This is because in an autoadvance playlist
+          // on Android devices with HLS renditions, the duration is incorrectly reported
+          // as infinity, due to which the content is not paused incorrectly
+          // So before we signal to the ad plugin , we want to pause content
+          cancelContentPlay(player);
           // signal to ad plugin that it's their opportunity to play a preroll
           player.trigger('readyforpreroll');
         }
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -266,12 +266,6 @@ const contribAdsPlugin = function(options) {
           player.ads.adTimeoutTimeout = window.setTimeout(function() {
             player.trigger('adtimeout');
           }, settings.prerollTimeout);
-          // We will try to pause content here. This is because in an autoadvance playlist
-          // on Android devices with HLS renditions, the duration is incorrectly reported
-          // as infinity, due to which the content is not paused incorrectly
-          // So before we signal to the ad plugin , we want to pause content
-          cancelContentPlay(player);
-          // signal to ad plugin that it's their opportunity to play a preroll
           player.trigger('readyforpreroll');
         }
       },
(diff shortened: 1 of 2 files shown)


src/plugin.js
```diff
@@ -261,6 +261,11 @@ const contribAdsPlugin = function(options) {
           player.ads.adTimeoutTimeout = window.setTimeout(function() {
             player.trigger('adtimeout');
           }, settings.prerollTimeout);
+          // We will try to pause content here. This is because in an autoadvance playlist
+          // on Android devices with HLS renditions, the duration is incorrectly reported
+          // as infinity, due to which the content is not paused incorrectly
+          // So before we signal to the ad plugin , we want to pause content
+          cancelContentPlay(player);
```
dmlap: Why does having a duration of infinity cause the content not to be paused as usual?

vdeshpande: We check for `this.duration() !== Infinity` on `adstart` and then we try to pause the player. But since we get a value of Infinity, the content is not paused.

dmlap: There's a pretty obvious follow-up question to that: why do we choose not to pause when duration is Infinity? The answer is that for many live video setups with client-side ads, we want the video to continue playing in the background so that we can be notified when the ad break is completing. This change would break those setups.

incompl: Let's add a comment that we should remove this workaround after this is in vjs: https://github.com/videojs/video.js/pull/3476

incompl: Ah I posted without seeing David's comment. We would have to check shouldPlayContentBehindAd here to cover that.

incompl: The reason it works is because content does not play behind ads on Android or iOS.

vdeshpande: Talking with Greg, since the playing of live video in playlist in the background is not a supported use-case, I'm going to go ahead with using shouldPlayContentBehindAd instead of cancelContentPlay as that would also address the concern of client side ads and live video in the background. 
We would also need to change or modify the adstart conditional check on Infinity duration in videojs-ima3.

src/plugin.js
```diff
@@ -261,7 +261,6 @@ const contribAdsPlugin = function(options) {
           player.ads.adTimeoutTimeout = window.setTimeout(function() {
             player.trigger('adtimeout');
           }, settings.prerollTimeout);
-          // signal to ad plugin that it's their opportunity to play a preroll
```
incompl: Was this comment removed because it was incorrect?

src/plugin.js
```diff
@@ -482,6 +481,12 @@ const contribAdsPlugin = function(options) {
           this.state = 'ad-playback';
         },
         contentupdate() {
+          // We know sources have changed, so we call CancelContentPlay
+          // to avoid playback of video in the background of an ad. Playback Occurs on
+          // an Android device if we do not call cancelContentPlay
```
incompl: Is it possible to add to this comment why only Android is affected?

dmlap:
@vdeshpande tests? Otherwise, looks good to me.

vdeshpande:
Thanks @dmlap ! I shall try and put in tests shortly

pagarwal123:
QA:LGTM

incompl:
@vdeshpande Can you rebase this so we can merge? Thanks!

vdeshpande:
@incompl Rebase done.

PR #198, opened 2016-10-21
Adding a new module to process metadata tracks.

ldayananda:


src/metadataTextTracks.js
```diff
@@ -0,0 +1,123 @@
+import videojs from 'video.js';
+
+const metadataTextTracks = {};
+let includedCues = {};
```
incompl: As it is this variable will be shared by all players, which I don't think you want. You'll need to make this a property of something player-specific such as the videojs-contrib-ads instance. Then your methods that manipulate this object will take the player as an argument, or something like that.

src/metadataTextTracks.js
```diff
@@ -0,0 +1,123 @@
...
+* Indicates that a cue has been used
+* @param cueId The Id associated with a cue
+**/
+const setCueIncluded = function(cueId) {
+  if (cueId !== undefined && cueId !== '') {
+    includedCues[cueId] = true;
+  }
+};
+
+module.exports = metadataTextTracks;
```
gkatsev: if you want to make this work ES6y, you'd want to change each function to be something like:
`export function setCueIncluded(cueId) ...` etc.

gkatsev: Then the import would be `import * as metadataTextTracks from './metadataTextTracks.js'`.

src/metadataTextTracks.js
```diff
@@ -0,0 +1,123 @@
...
+  } else {
+    tracks.addEventListener('addtrack', (event) => {
+      const track = event.track;
+      prepareTrack(track);
+    });
+  }
+};
+
+/**
+* @override only if necessary
```
incompl: I'm not sure the override messages are useful, my inclination would be to leave them out.

incompl: Or, describe the conditions under which you'd need to override it?

src/metadataTextTracks.js
```diff
@@ -0,0 +1,123 @@
+import videojs from 'video.js';
```
incompl: What do you think of renaming this file something like `adCueTextTracks.js` to make it more clear what the feature is for? Also, a comment at the top of the file describing what the feature is for would be useful, I've been adding one to each of the feature files.

src/metadataTextTracks.js
ldayananda: This is a very, very good point. I'll look into making this player specific.

src/metadataTextTracks.js
ldayananda: I mostly did it this way to be consistent with 'snapshot.js'. I do agree that it would be better for this to be ES6 compatible, but if we do so it should be across both files.

incompl: All of the modules do this I think, so if we change them let's change them all

src/metadataTextTracks.js
ldayananda: Sure, I could move this below the explanation of default behavior

src/metadataTextTracks.js
ldayananda: I'll add a descriptive comment at the top of the file.

The main reason for me going with "metadataTextTracks" was because there are two main methods in this file: process(on general metadata text tracks) and processAdTrack(on specifically ad text tracks). Any suggestions that encompass both would be appreciated! Perhaps `cueTextTracks.js`?

incompl: My goal for the suggestion is to let someone new to the project understand why this feature is useful. `cueTextTracks.js`is closer to that goal in my opinion, because "cues" are what the feature is intended to support.

incompl: On the other hand, I wouldn't mind going even further with a name like `adCues.js`, since that fully communicates the goal, even if the implementation is technically not so specific.

src/metadataTextTracks.js
ldayananda: I will leave it like this for this particular PR and open another that exports each function instead across all the modules.

src/metadataTextTracks.js
```diff
@@ -0,0 +1,128 @@
...
+      track.mode = 'hidden';
+      processTrack(player, track);
+    }
+  };
+
+  // Text tracks are available
+  if (tracks.length > 0) {
+    for (let i = 0; i < tracks.length; i++) {
+      const track = tracks[i];
+      prepareTrack(track);
```
gkatsev: This calls `prepareTrack` for each text track on the player and it in turn enables each one. However, we don't necessarily want to enable all the metadata text tracks. There may be old ones around we don't want enabled.

ldayananda: So if I understand correctly, if a track is 'disabled' at this point, there should be no reason to set it to 'hidden'?

gkatsev: Not necessarily. However, some tracks could be disabled but shouldn't be re-enabled while some might be ok to re-enable. This is particularly an issue because if you add a text track via `addTextTrack` you are unable to remove it, so, the only thing you can do (beyond removing the video element and creating a new one) is to just disable the track.

May not necessarily be something that we handle in code but maybe only in documentation. But it's important to know so we can make an informed decision.

README.md
```diff
@@ -99,6 +99,37 @@ player.ads.ad = {
...
+    player.ads.metadataTextTracks.processAdTrack(player, cues, processCue, cancelAds);
+  });
+}
+```
+
+For more information on the utility methods that are available, see [metadataTextTracks.js](https://github.com/videojs/videojs-contrib-ads/blob/master/src/metadataTextTracks.js).
+
+### setTrackMode
+
+A track is 'enabled' if the track.mode is set to `hidden` or `showing`. Otherwise, a track is `disabled` and is not updated. It is important to note that some tracks may be disabled as a workaround of not being able to remove them, and so should not be re-enabled. Ad integrations should be careful about setting the mode of tracks in these cases and shadow `setTrackMode` to determine which tracks are safe to change.
```
incompl: Maybe add an example of how setTrackMode is used?

README.md
```diff
@@ -99,6 +99,37 @@ player.ads.ad = {
...
+
+```
+function processTrack(player, track) {
+  track.addEventListener('cuechange', function() {
+    var cues = this.cues;
+    var processCue = function() {
+      // Make an ad request
+      ...
+    };
+    var cancelAds = function() { ... };
```
ldayananda: Will do

src/metadataTextTracks.js
```diff
@@ -41,13 +41,11 @@ metadataTextTracks.process = function(player, processTrack) {
...
 * @see https://github.com/videojs/video.js/blob/master/docs/guides/text-tracks.md
-* Default behavior is to set the track to hidden, @override if this is not desired
+* Default behavior is to do nothing, @override if this is not desired
 * @param track The text track to set the mode on
 */
 metadataTextTracks.setTrackMode = function(track) {
-  // Make sure track is 'hidden' rather than 'disabled' as it may have already
-  // existed and had been 'disabled' during a playlist change
-  track.mode = 'hidden';
+  return;
```
ldayananda: Updated this from `track.mode = 'hidden'` to avoid changing the mode of text tracks restored by the snapshot. Integrations should define their own behaviour.

ldayananda:
The module has been renamed to cueTextTracks.

Also, I have changed the default behavior of `setTrackMode` to not modify the mode to avoid changing the mode of textTracks restored from the snapshot as @gkatsev suggested.

incompl:
LGTM

README.md
```diff
@@ -164,6 +164,49 @@ A macro such as {pageVariable.foobar} allows the user access the value of any pr
 | Undefined | Logs warning and returns empty string |
 | Other     | Logs warning and returns empty string |
 
+## Cue Metadata Tracks
+
+An optional feature that allows the manipulation of metadata tracks, specifically in the case of working with advertising cue points.
+
+For example, an ad integration may want to make an ad request when a cuepoint change has been observed. To do this, an ad integration would need to do somthing like this:
```
incompl: somthing -> something

package.json
```diff
@@ -6,7 +6,7 @@
     "type": "git",
     "url": "https://github.com/videojs/videojs-contrib-ads.git"
   },
-  "version": "4.1.5",
+  "version": "4.1.6-0",
```
incompl: Let's not check in the version number change

src/cueTextTracks.js
```diff
@@ -0,0 +1,136 @@
...
+  const tracks = player.textTracks();
+  const prepareTrack = function(track) {
+    if (track.kind === 'metadata') {
+      cueTextTracks.setTrackMode(track);
+      processTrack(player, track);
+    }
+  };
+
+  // Text tracks are available
+  if (tracks.length > 0) {
```
incompl: Do you need this condition? Would it be safer to both run the forloop and attach the event handler?

src/cueTextTracks.js
```diff
@@ -0,0 +1,136 @@
...
+/**
+* This feature allows metadata text tracks to be manipulated once they are available,
+* usually after the 'loadstart' event is observed on the player
+* @param player A reference to a player
+* @param processTrack A method that performs some operations on a metadata text track
+**/
+cueTextTracks.process = function(player, processTrack) {
+  const tracks = player.textTracks();
+  const prepareTrack = function(track) {
+    if (track.kind === 'metadata') {
```
incompl: Is it possible for there to be metadata tracks we don't want to set the mode of?

incompl: Oh I guess we're only here if they deliberately selected this track

src/cueTextTracks.js
```diff
@@ -0,0 +1,136 @@
...
+* @param processCue A method that uses a cue to make some ad request in the ad implementation
+* @param [cancelAds] A method that dynamically cancels ads in the ad implementation
+**/
+cueTextTracks.processAdTrack = function(player, cues, processCue, cancelAds) {
+  player.ads.includedCues = {};
+
+  // loop over set of cues
+  for (let i = 0; i < cues.length; i++) {
+    const cue = cues[i];
+    let cueData, cueId, startTime;
```
incompl: By convention I believe we're doing one let per line, does vjsstandard accept this?

README.md
```diff
@@ -164,6 +164,49 @@ A macro such as {pageVariable.foobar} allows the user access the value of any pr
...
+
+`player.ads.cueTextTracks.process(player, processTrack)`
```
incompl: Here are the naming changes we came up with in our discussion just now:

process      -> processMetadataTracks*
prepareTrack -> setModeAndProcess
  setTrackMode -> setMetadataTrackMode*
  processMetadataTrack -> processMetadataTrack*

src/cueTextTracks.js
```diff
@@ -0,0 +1,136 @@
...
+      videojs.log.warn('Skipping as this is not a supported ad cue.', cue);
+      return;
+
+    // Continue processing supported cue
+    } else {
+      cueId = this.getCueId(player, cue);
+      startTime = cue.startTime;
+
+      // Skip ad if cue was already used
+      if (cueIncluded(player, cueId)) {
```
incompl: This is another one we might rename to something like cueAlreadyUsed

incompl: or cueAlreadyProcessed, cueAlreadySeen, something that indicates "what it's included in"

README.md
ldayananda: oops

package.json
ldayananda: Yes, this needs to be rebased soon

src/cueTextTracks.js
ldayananda: It hasn't failed me but I'll change it to follow convention

src/cueTextTracks.js
ldayananda: good point

test/videojs.ads.snapshot.test.js
```diff
@@ -411,10 +411,28 @@ QUnit.test('No snapshot if duration is Infinity', function(assert) {
...
+    'bcls/captions/adding_captions_to_videos_french.vtt';
+  const originalAddTrack = this.player.addTextTrack;
+  const originalTextTracks = this.player.textTracks;
+  let mockTracks = [];
 
   // No text tracks at start
   assert.equal(this.player.remoteTextTracks().length, 0);
+  assert.equal(this.player.textTracks().length, 0);
+
+  // This is mocked because of native text track behavior
```
incompl: Maybe add what behavior you mean to the comment?

ldayananda: Updated

src/cueTextTracks.js
ldayananda: There is now a comment above the else condition that explains this

incompl:
Looked good last I looked, if all that changed is incorporating feedback I'm LGTM

PR #200, opened 2016-10-27
contentplayback is contentplay-back

incompl:


src/plugin.js
```diff
@@ -464,6 +464,16 @@ const contribAdsPlugin = function(options) {
           window.clearTimeout(player.ads.cancelPlayTimeout);
           player.ads.cancelPlayTimeout = null;
         }
+
+        // This was removed because now that "playing" is fixed to only play after
+        // preroll, any integration should just use the "playing" event. However,
+        // we found out some 3rd party code relied on this event, so we've added
+        // it back in to give people more time to update their code.
```
ldayananda: Perhaps add `temporarily` to make it obvious this is going away at some point?

ldayananda:
Last thing would be to update the 4.0 migration guide and documentation related to this to indicate deprecation rather than removal.

Otherwise, LGTM

incompl:
Technically it was removed in 4.0.0 and will be re-added in a later version
so I want to leave it in there. It feels dishonest to say we deprecated it
after the fact.

ldayananda:
True, perhaps it will cause less confusion if we specify the exact versions in which it was removed and re-added. That is, unless this change is going to be a major update.

I still feel the README should be updated to reflect this change.

incompl:
I'll add it to the readme as a warning to not use it

bcvio:
LGTM

pagarwal123:
QA: LGTM

gkatsev:
LGTM

PR #202, opened 2016-11-02
Fix text track support in snapshot

incompl:


src/plugin.js
```diff
@@ -206,6 +206,9 @@ const contribAdsPlugin = function(options) {
 
   player.ads.adMacroReplacement = adMacroReplacement.bind(player);
 
+  // Expose snapshot tools for testing
+  player.ads._snapshot = snapshot;
```
ldayananda: Why is this not already available to tests via player.ads.snapshot? Can we import it into the test file instead?

ldayananda: What I meant was: is it possible to import the snapshot module into the test file?

ldayananda:
LGTM

pagarwal123:
QA-LGTM

PR #205, opened 2016-11-10
Add import of videojs in cancelContentPlay.js

Ambroos:
This import was probably forgotten. In most situations where videojs is available in the global scope this wouldn't have resulted in any issues, but when packaging everything with Webpack or Browserify this is not always the case.

Because of this bug using videojs-contrib-ads with Webpack currently breaks. A patch release would be much appreciated!


incompl:
LGTM

incompl:
We're doing a release soon that will include this.

Ambroos:
Thanks a lot!

PR #207, opened 2016-11-10
blank div to prevent content flash covers ad on iPad

marguinbc:
Description: In cancelContentPlay, there is code to hide the video element and overlay a black filled div over the video player until 'adplaying', 'adtimeout', 'adserror', 'adscanceled', 'adskip', or 'playing' fires at which point the div is hidden and the video element made visible.

Existing code: This was happening in IMA3 and FW and in Chrome or Safari on the iPad. This code currently tests whether or not to run by checking videojs.browser.IS_IOS.

Proposed change: This is not sufficient as we also need to make sure it is NOT an iPad. This fix modifies the check to test for videojs.browser.IS_IOS && !videojs.browser.IS_IPAD.

Expected behavior:
1. Load video with preroll ad in browser on iPad
2. Preroll ad loads and plays, ad is visible, audio is audible
3. Content plays

Actual behavior:
1. Load video with preroll ad in browser on iPad
2. Preroll ad loads and plays, ad is NOT visible, audio is audible
3. Content plays

Tested this using the current versions of Freewheel AND IMA3 on:

iPad iOS 10.1.1 with Safari
iPad iOS 10.1.1 with Chrome 54.0.2840.91

iPhone iOS 10.1.1 with Safari*
iPhone iOS 10.1.1 with Chrome

Desktop Mac OS X El Capitan Safari  10.0.1 (11602.2.14.0.7)
Desktop Mac OS X El Capitan Chrome 54.0.2840.71

NOTE: On IMA3 on the iPhone, I see a content flash/poster before the video plays, which is weird because the code should run on iPhone since IS_IOS and !IS_IPAD should both be true.  TESTING with the original player on iPhone also shows this behavior though so it does not appear to be added by this change.  The original test player DOES have the black div above the player in iPad though.

### DIFF
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -14,7 +14,7 @@ const cancelContentPlay = function(player) {
   }
 
   // Avoid content flash on non-iPad iOS
-  if (videojs.browser.IS_IOS) {
+  if ((videojs.browser.IS_IOS) && (!videojs.browser.IS_IPAD)) {
 
     const width = player.currentWidth ? player.currentWidth() : player.width();
     const height = player.currentHeight ? player.currentHeight() : player.height();
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -14,7 +14,7 @@ const cancelContentPlay = function(player) {
   }
 
   // Avoid content flash on non-iPad iOS
-  if ((videojs.browser.IS_IOS) && (!videojs.browser.IS_IPAD)) {
+  if (videojs.browser.IS_IOS && videojs.browser.IS_IPHONE) {
 
     const width = player.currentWidth ? player.currentWidth() : player.width();
     const height = player.currentHeight ? player.currentHeight() : player.height();


src/cancelContentPlay.js
```diff
@@ -14,7 +14,7 @@ const cancelContentPlay = function(player) {
   }
 
   // Avoid content flash on non-iPad iOS
-  if (videojs.browser.IS_IOS) {
+  if ((videojs.browser.IS_IOS) && (!videojs.browser.IS_IPAD)) {
```
ldayananda: I believe there is an IS_IPHONE option you can use instead https://github.com/videojs/video.js/blob/3a859f97d0647fc3d1ef6b2f4b755c71da8bef63/src/js/utils/browser.js#L23

marguinbc: Thanks! I did see it but since the comment said non iPad, I was trying to match the logic to the comment.  Changed it to IS_IPHONE still works on iPad and doesn't break iPhone in IMA3 or FW, so I'll resubmit.

ldayananda:
LGTM

marguinbc:
QA prerelease: videojs-contrib-ads@4.2.0-3

pagarwal123:
for ima3: @brightcove/videojs-ima3@2.15.3-1
http://cs1.brightcodes.net/pdias/samples/placeholder-essence-display-none.html: (mentioned in BC-37210) - looks good for ipads but the issue is still seen in iphone 10 (setting playsinline)
Reproduce the iphone 10 issue with: http://solutions.brightcove.com/pagarwal/Black_Div_bc37447/bc37210.html

http://cs1.brightcodes.net/tzahora/jira/standalones/bc/37447/index.html (mentioned in BC- 37447) - issue was seen in ipad 8 and ipad 10. looks good for ipad 8. For ipad 10: player is not bumped down, but the ad does not play. 
Reproduce the ipad 10 issue with : 
http://solutions.brightcove.com/pagarwal/Black_Div_bc37447/bc37447.html

for freewheel: @brightcove/videojs-freewheel@2.5.0-3
looks good for ipads but issue still seen in ios 10 (set playsinline)

pagarwal123:
BC-37210- Tested on ipads 8, 9, 10 and issue : 'player is shifted below and a blank space appears temporarily on top of it' is resolved link: http://solutions.brightcove.com/pagarwal/Black_Div_bc37447/bc37210.html

There is a different issue with iphones (black div after exiting fullscreen) and iphone 10 playsinline issue which is being tracked by other jira: https://jira.brightcove.com/browse/BC-37454

There is a different timeout issue for http://solutions.brightcove.com/pagarwal/Black_Div_bc37447/bc37447.html for ipad 10 which is being tracked by jira: https://jira.brightcove.com/browse/BC- 37447

pagarwal123:
QA:LGTM

incompl:
Approved for deploy to production

PR #212, opened 2016-11-14
Fix to prevent clickthrough when progress bar is clicked

vdeshpande:
**Proposal:**
This is a proposal fix to prevent the click through from happening when user clicks on the progress control.

**Current situation:**
What currently happens when user clicks on the progress control is when the user clicks on the progress control, it takes you to the ad-clickthrough page.

**Fix proposal:**
What should happen when user clicks on the progress control - nothing.

I have tested this across multiple platforms, and devices and the ad-clickthrough seems to be prevented with this fix.

### DIFF
--- a/src/plugin.scss
+++ b/src/plugin.scss
@@ -4,7 +4,7 @@ Styles for ad integrations.
 
 // Prevent seeking during ad playback
 .vjs-ad-playing.vjs-ad-playing .vjs-progress-control {
-  pointer-events: none;
+  pointer-events: auto;
 }
 
 // Make the progress bar yellow for ads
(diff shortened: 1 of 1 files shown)


dmlap:
Based on the comment above the changed rule, this change would allow seeking during ads. How is seeking prevented with this change in place?

vdeshpande:
Speaking with @dmlap , I'm going to try and see if I can come up with something else.

The fix does work, however it gives the user permission to interact with the scrub bar as much as they want (FYI: Seeking does not occur during ad-playback)
We do not want the ability of setting pointer-events to none to be dropped, so a different approach might be the better option to pursue

vdeshpande:
@dmlap How does the above change look ?

incompl:
Replaced by #219

PR #213, opened 2016-11-18
Try to set correct currentTime after postroll when using HLS

incompl:


incompl:
We decided this is too much of a hack and we need a better solution

PR #216, opened 2016-11-22
Ended event after seeking backwards, plus don't use snapshot as global variable

incompl:


incompl:
Built on #214

src/plugin.js
```diff
@@ -526,18 +536,20 @@ const contribAdsPlugin = function(options) {
           } else {
             this.state = 'ads-ready?';
           }
-          // When a new source is loaded into the player, we should remove the snapshot
-          // to avoid confusing player state with the new content's state
-          // i.e When new content is set, the player should fire the ended event
-          if (this.snapshot && this.snapshot.ended) {
-            this.snapshot = null;
-          }
```
ldayananda: What if a contentupdate occurs before the end of a video? i.e in a playlist

incompl: `contentupdate` should happen any time the source changes to a new source. With the new logic snapshot is nulled on contentupdate in any state, rather than just in `content-playback`, so that's the big change here.

ldayananda: 👍

src/plugin.js
```diff
@@ -209,6 +216,9 @@ const contribAdsPlugin = function(options) {
   // Start sending contentupdate events for this player
   initializeContentupdate(player);
 
+  // Global contentupdate handler for resetting plugin state
+  player.on('contentupdate', player.ads.initialize);
```
bcvio: Since this is a recurring event,  renaming `player.ads.initialize` to `player.ads.reset` feels more natural to me

bcvio:
LGTM

pagarwal123:
QA LGTM

PR #218, opened 2016-11-25
Do not throw error if there is no snapshot

### LINKED ISSUE #217: There is no snapshot on live streams, no need to throw error
`You cannot use videoElementRecycled while there is no snapshot.` error is thrown dozens of time in case of live streams and clutter the browser console.

mkhazov:
fixes #217

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -170,7 +170,7 @@ const contribAdsPlugin = function(options) {
       let currentSrcChanged;
 
       if (player.ads.shouldPlayContentBehindAd(player)) {
-        return true;
+        return false;
       }
       if (!this.snapshot) {
         throw new Error(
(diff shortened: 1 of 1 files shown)


incompl:
I have a proposed improvement: the purpose of the `videoElementRecycled`function is to determine if the video element is recycled for ad playback in this particular configuration. If there is no snapshot, the current answer is "I don't know", hence the error. In your case, you could take advantage of the fact that you know the video element is not recycled during a live stream, and make that explicit logic. Something like:

```
if player.duration is infinity
  return true
if !snapshot
  throw error
```

That way, in a case where the video element is recycled but there is no snapshot currently, it won't erroneously report false.

incompl:
Woops I had it reversed in my pseudocode: it should return false, not true. If the content is playing back behind the ad, the video element is not recycled for ad playback. Sorry about that.

mkhazov:
If return false, player.play() is not called, so on adend the playback stays paused.
Probably some refactoring is needed here.

incompl:
@mkhazov Do you want this re-reviewed or are you still working on it?

incompl:
I believe this was done in https://github.com/videojs/videojs-contrib-ads/pull/259, it's worth checking if this issue is fixed now. Feel free to reopen the PR if there is still something to be done here. Just cleaning up the project a bit.

PR #219, opened 2016-11-29
Prevent clickthrough when progress bar clicked during ad

incompl:
Based on #212

### DIFF
--- a/src/plugin.scss
+++ b/src/plugin.scss
@@ -3,12 +3,15 @@ Styles for ad integrations.
 */
 
 // Prevent seeking during ad playback
-.vjs-ad-playing.vjs-ad-playing .vjs-progress-control::after {
-  pointer-events: auto;
+.vjs-ad-playing.vjs-ad-playing .vjs-progress-control:after {
   width: 100%;
   height: 100%;
   position: absolute;
-  content: "\000a0"
+  content: " ";
+  user-select: none;
+  -ms-user-select: none;
+  -moz-user-select: none;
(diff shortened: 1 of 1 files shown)


src/plugin.scss
```diff
@@ -3,8 +3,12 @@ Styles for ad integrations.
 */
 
 // Prevent seeking during ad playback
-.vjs-ad-playing.vjs-ad-playing .vjs-progress-control {
-  pointer-events: none;
+.vjs-ad-playing.vjs-ad-playing .vjs-progress-control::after {
```
gkatsev: Does sass transform the double colon (`::`) into a single colon (`:`)? IE8 only supports a single colon.

src/plugin.scss
```diff
@@ -3,8 +3,12 @@ Styles for ad integrations.
...
+  pointer-events: auto;
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  content: "\000a0"
```
gkatsev: Would it be better to just use a space?

src/plugin.scss
gkatsev: We should potentially disallow user selection on this element.

src/plugin.scss
incompl: I'm fine with space, just copied this from a previous version

src/plugin.scss
incompl: Nope it doesn't, good catch

pagarwal123:
QA-LGTM

PR #220, opened 2016-12-06
Fix placehoder div on ios10 playsinline

marguinbc:
This fix adds a check for playsinline on iPhone to determine whether to add the black div used to cover the content player and prevent flash of content/poster.  Without this fix, a player with playsinline on an iPhone with iOS10 shows the black div over the player preventing the ad from being viewed.  

The black div would also appear on iPhones on iOS10 without playsinline - in this case, the div remains visible after returning from fullscreen during a paused ad and before content begins playing. Added a check for playsinline and the placeholder on fullscreenchange to remove the div and show the player.

To test scenario 1: 

- setup player with playsinline and ad plugin (ima3 or freewhee)
- play video on iPhone with iOS 10
Before fix: black div covers ad
After fix: black div is never presented over ad before playing ad

Risk: potential to have content-flash when playing inline without the div to cover the poster/content player

To test scenario 2: 

- setup player without playsinline
- play content fullscreen on iPhone with iOS 10
- during ad pause and/or pinch to close ad player
Before fix: black div covers ad when returning from fullscreen
After fix: div is hidden and content player begins content when ad is complete

### DIFF
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -32,16 +32,16 @@ const cancelContentPlay = function(player) {
     player.el_.style.display = 'none';
 
     // Unhide the player and remove the placeholder once we're ready to move on.
-    player.one(['adplaying', 'adtimeout', 'adserror', 'adscanceled', 'adskip',
+    player.one(['adstart', 'adplaying', 'adtimeout', 'adserror', 'adscanceled', 'adskip',
                 'playing'], function() {
       player.el_.style.display = 'block';
       placeholder.remove();
     });
 
-    // Detect fullscreen change, remove placeholder and show player.
-    // On iOS 10 Safari is supposed to also add playsinline to the player
+    // Detect fullscreen change, if returning from fullscreen and placeholder exists,
+    // remove placeholder and show player whether or not playsinline was attached.
     player.on('fullscreenchange', function () {
-      if (placeholder) {
+      if (placeholder && player.hasClass('vjs-fullscreen')) {
         player.el_.style.display = 'block';
         placeholder.remove();
       }
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -38,10 +38,10 @@ const cancelContentPlay = function(player) {
       placeholder.remove();
     });
 
-    // On iOS 10 on iPhone, playsinline is added when pinching to return from
-    // fullscreen. Detect fullscreen change, remove placeholder and show player.
+    // Detect fullscreen change, remove placeholder and show player.
+    // On iOS 10 Safari is supposed to also add playsinline to the player
     player.on('fullscreenchange', function () {
-      if (player.el_.hasAttribute('playsinline') && placeholder) {
+      if (placeholder) {
         player.el_.style.display = 'block';
         placeholder.remove();
       }
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -37,6 +37,16 @@ const cancelContentPlay = function(player) {
       player.el_.style.display = 'block';
       placeholder.remove();
     });
+
+    // On iOS 10 on iPhone, playsinline is added when pinching to return from
+    // fullscreen. Detect fullscreen change, remove placeholder and show player.
+    player.on('fullscreenchange', function () {
+      if (player.el_.hasAttribute('playsinline') && placehoder) {
+        player.el_.style.display = 'block';
+        placeholder.remove();
+      }
+    })
+
   }
 
   // The timeout is necessary because pausing a video element while processing a `play`
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -13,7 +13,7 @@ const cancelContentPlay = function(player) {
     return;
   }
 
-  // Avoid content flash on non-iPad iOS
+  // Avoid content flash on non-iPad iOS and iPhones on iOS10 with playsinline
   if ((videojs.browser.IS_IOS && videojs.browser.IS_IPHONE) && !player.el_.hasAttribute('playsinline')) {
 
     const width = player.currentWidth ? player.currentWidth() : player.width();
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -14,7 +14,7 @@ const cancelContentPlay = function(player) {
   }
 
   // Avoid content flash on non-iPad iOS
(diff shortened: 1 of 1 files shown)


src/cancelContentPlay.js
```diff
@@ -37,6 +37,16 @@ const cancelContentPlay = function(player) {
       player.el_.style.display = 'block';
       placeholder.remove();
     });
+
+    // On iOS 10 on iPhone, playsinline is added when pinching to return from
+    // fullscreen. Detect fullscreen change, remove placeholder and show player.
+    player.on('fullscreenchange', function () {
+      if (player.el_.hasAttribute('playsinline') && placehoder) {
```
gkatsev: typo, `placeho*l*der`

src/cancelContentPlay.js
```diff
@@ -14,7 +14,7 @@ const cancelContentPlay = function(player) {
   }
 
   // Avoid content flash on non-iPad iOS
-  if (videojs.browser.IS_IOS && videojs.browser.IS_IPHONE) {
+  if ((videojs.browser.IS_IOS && videojs.browser.IS_IPHONE) && (!player.el_.hasAttribute('playsinline'))) {
```
gkatsev: I don't think that the inner parens are necessary, at least on the `hasAttribute` part.

marguinbc: Thanks for the comments @gkatsev. Fixed code based on both suggestions.

src/cancelContentPlay.js
```diff
@@ -14,7 +14,7 @@ const cancelContentPlay = function(player) {
   }
 
   // Avoid content flash on non-iPad iOS
-  if (videojs.browser.IS_IOS && videojs.browser.IS_IPHONE) {
+  if ((videojs.browser.IS_IOS && videojs.browser.IS_IPHONE) && !player.el_.hasAttribute('playsinline')) {
```
incompl: Minor, but can you update the comment above this line?

marguinbc: done :)

incompl:
LGTM

marguinbc:
prerelease for testing is videojs-contrib-ads@4.1.5-0

src/cancelContentPlay.js
```diff
@@ -37,6 +37,16 @@ const cancelContentPlay = function(player) {
       player.el_.style.display = 'block';
       placeholder.remove();
     });
+
+    // On iOS 10 on iPhone, playsinline is added when pinching to return from
+    // fullscreen. Detect fullscreen change, remove placeholder and show player.
+    player.on('fullscreenchange', function () {
+      if (player.el_.hasAttribute('playsinline') && placeholder) {
+        player.el_.style.display = 'block';
```
gkatsev: this looks like the same like line 37. We should consider pulling it out into a function. Maybe even have two
* hide player and enable placeholder
* show player and disable placeholder

Doesn't have to change for this PR just something I noticed.

src/cancelContentPlay.js
```diff
@@ -37,6 +37,16 @@ const cancelContentPlay = function(player) {
       player.el_.style.display = 'block';
       placeholder.remove();
     });
+
+    // On iOS 10 on iPhone, playsinline is added when pinching to return from
+    // fullscreen. Detect fullscreen change, remove placeholder and show player.
+    player.on('fullscreenchange', function () {
+      if (player.el_.hasAttribute('playsinline') && placeholder) {
```
pagarwal123: Just noticed something. Can we also have a check over here: `!player.el_.hasAttribute('playsinline')` so that on fullscreen change black div is not seen even in the case if `playsinline` is not set (in case of iphone 8 and 9)? Does that sound ok?

marguinbc: Are you seeing the black div when returning from fullscreen? This line says if playsinline and the black div is there - hide it. As I understand it, when returning from fullscreen with pinch-to-close, Safari adds playsinline. 

How are you exiting fullscreen? Pinch-to-close, Done or Fullscreen button?

marguinbc:
Created an IMA3 prerelease with this contrib-ads fix: @brightcove/videojs-ima3@2.15.4-5

src/cancelContentPlay.js
pagarwal123: @marguinbc For iphone 10 - Tested both scenarios as suggested above:
`Scenario 1: setup player with playsinline and ad plugin (ima3 or freewheel)` 👍 
- I dont see a black div : case1 - Preroll is playing 👍 
                                          case2 - Returning from fullscreen 👍 

`Scenario 2: setup player without playsinline`
-I see a black div when returning from fullscreen

Checked by exiting fullscreen in all the 3 ways: Pinch-to-close, Done or Fullscreen button

src/cancelContentPlay.js
```diff
@@ -37,6 +37,16 @@ const cancelContentPlay = function(player) {
       player.el_.style.display = 'block';
       placeholder.remove();
     });
+
+    // Detect fullscreen change, remove placeholder and show player.
+    // On iOS 10 Safari is supposed to also add playsinline to the player
```
gkatsev: For correctness, I'm mentioning it here, not sure whether we need to update the player. iOS 10 would only add `playsinline` if you pinch to exit or if you hit the exit fullscreen button on the bottom right side of the player. Hitting "done" should not be adding this property.
Also, it adds the `controls` attribute in addition to `playsinline`, we might need to account for that as well.

pagarwal123:
QA:LGTM

src/cancelContentPlay.js
```diff
@@ -32,11 +32,21 @@ const cancelContentPlay = function(player) {
     player.el_.style.display = 'none';
 
     // Unhide the player and remove the placeholder once we're ready to move on.
-    player.one(['adplaying', 'adtimeout', 'adserror', 'adscanceled', 'adskip',
+    player.one(['adstart', 'adplaying', 'adtimeout', 'adserror', 'adscanceled', 'adskip',
```
incompl: Why is adstart added in addition to adplaying?

marguinbc: I was originally trying to hide the black div using fullscreenschange but in one scenario (I think it was with a specific ad plugin) fullscreenchange wasn't firing and hiding the div.  Additionally, it wasn't being removed in the player.one('adplaying') case so I used adstart. The risk is removing the div too soon and allowing content flash on iOS without playsinline.

PR #226, opened 2016-12-30
Re-fix iOS content flash

incompl:

PR #230, opened 2017-01-04
Placeholder div is jumping down

nidin:
Since default position is relative, placeholder div is jumping down


incompl:
I believe this issue is resolved by #226. I tested this PR with the changes in #226 and they did not seem to be compatible in the test page I was using. Would you mind trying out #226 and seeing if it solves your issue?

incompl:
Assuming this issue is resolved, please let us know if not!

PR #231, opened 2017-01-10
prevent ios10 preroll autoplay playsinline without muted

marguinbc:
On iOS 10, if the player has the playsinline attribute and is configured to autoplay, but is not muted, we currently try to autoplay the ad. In content-set -> play, we cancelContentPlay and hide the native poster to prepare for playback - however the native browser on IOS prevents the ad from loading.

This results in a loading spinner for the duration of the ad or until the ad times out at which point the player presents the big play button and poster for the content.

This fix adds a new function cancelAutoplayAdOnIOS which will check to see if the player is on an iPhone, has autoplay, playsinline and muted attributes - if so, we remove the autoplay attribute/setting, pause the player and return true to indicate canceling the autoplay of the ad.

Otherwise, we simply perform the original tasks of setting the state to ads-ready?, call cancelContentPlay and removeNativePoster.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -214,23 +214,22 @@ const contribAdsPlugin = function(options) {
     // to autoplay the ad.  Also pauses the player if it has started playing and
     // removes the autoplay attribute or player setting.
     cancelAutoplayAdOnIOS(somePlayer) {
-      if( videojs.IS_IPHONE && player.el_.hasAttribute('playsinline')
-        && (player.autoplay() || player.el_.hasAttribute('autoplay'))
-        && (!player.muted() || !player.el_.hasAttribute('muted'))) {
+      if (videojs.IS_IPHONE && player.el_.hasAttribute('playsinline') &&
+        (player.autoplay() || player.el_.hasAttribute('autoplay')) &&
+        (!player.muted() || !player.el_.hasAttribute('muted'))) {
 
-          if (!player.paused()) {
-            player.pause();
-          }
-          if (player.autoplay() === true) {
-            player.autoplay(false);
-          }
-          if (player.el_.hasAttribute('autoplay')) {
-            player.el_.removeAttribute('autoplay');
-          }
+        if (!player.paused()) {
+          player.pause();
+        }
+        if (player.autoplay() === true) {
+          player.autoplay(false);
+        }
+        if (player.el_.hasAttribute('autoplay')) {
+          player.el_.removeAttribute('autoplay');
+        }
         return true;
-      } else {
-        return false;
       }
+      return false;
     }
   };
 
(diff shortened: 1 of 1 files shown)


src/plugin.js
```diff
@@ -230,10 +254,12 @@ const contribAdsPlugin = function(options) {
           this.state = 'ads-ready';
         },
         play() {
```
incompl: Is this possible to be a manual play rather than an autoplay?

marguinbc: @incompl , from what I can tell in testing, the only time we are getting a play event while in the content-set state is when initially loading and autoplaying a video. I tried everything I can with playlists and play always comes when we are in the ads-ready state....

incompl:
LGTM

src/plugin.js
```diff
@@ -206,8 +206,31 @@ const contribAdsPlugin = function(options) {
...
+    // and muted attributes or player settings. Otherwise, we should attempt
+    // to autoplay the ad.  Also pauses the player if it has started playing and
+    // removes the autoplay attribute or player setting.
+    cancelAutoplayAdOnIOS(somePlayer) {
+      if (videojs.IS_IPHONE && player.el_.hasAttribute('playsinline') &&
+        (player.autoplay() || player.el_.hasAttribute('autoplay')) &&
+        (!player.muted() || !player.el_.hasAttribute('muted'))) {
+
+        if (!player.paused()) {
+          player.pause();
```
incompl: This is a side effect of this boolean check, we should move this out into an else block.

src/plugin.js
```diff
@@ -206,8 +206,31 @@ const contribAdsPlugin = function(options) {
...
 
+    // Return true if the ad should NOT autoplay on iPhone on iOS 10 and above.
+    // This is only true if the player does not have autoplay, playsinline
+    // and muted attributes or player settings. Otherwise, we should attempt
+    // to autoplay the ad.  Also pauses the player if it has started playing and
+    // removes the autoplay attribute or player setting.
+    cancelAutoplayAdOnIOS(somePlayer) {
+      if (videojs.IS_IPHONE && player.el_.hasAttribute('playsinline') &&
+        (player.autoplay() || player.el_.hasAttribute('autoplay')) &&
+        (!player.muted() || !player.el_.hasAttribute('muted'))) {
```
incompl: Let's add a link to this issue so people remember that we'd like to simplify this in the future if possible: https://github.com/videojs/video.js/issues/3970

marguinbc:
Freewheel prerelease: @brightcove/videojs-freewheel@2.6.2-0

IMA3 prerelease:  @brightcove/videojs-ima3@2.16.1-1

incompl:
Replaced by the feature on SVT that removes the autoplay attribute on the video element if playsonline and muted aren't there on iOS10.

PR #232, opened 2017-01-13
Do not take a snapshot when we already know there will not be a postroll

ldayananda:


incompl:
As a note, it used to be like this and we rolled it back.

https://github.com/videojs/videojs-contrib-ads/commit/abe17da36a55a3c39d36a8be8e8f189b5e0a0e11

I'm not sure what issue this caused. One path forward would be to confirm the fixes that were made in the next release after that one: https://github.com/videojs/videojs-contrib-ads/blob/master/CHANGELOG.md#332

ldayananda:
Pre-release for QA: videojs-contrib-ads@4.2.1-0

@incompl I'm a little confused, what do you mean by "confirm the fixes"?

incompl:
In the commit I linked, we rolled back this change. The CHANGELOG for 3.3.2 shows what bugs were fixed by the rollback. If we are going to add this change back in, we should verify that the things that were fixed by rolling it back are still fixed.

apadhye:
QA LGTM

PR #233, opened 2017-01-13
Clean up tests, add a redispatch unit test

incompl:


scripts/server.js
```diff
@@ -19,7 +19,7 @@ const nameify = (str) =>
 const srces = {
   css: 'src/plugin.scss',
   js: 'src/plugin.js',
-  tests: glob.sync('test/**/*.test.js')
+  tests: glob.sync('test/**/test.*.js')
```
incompl: This is to have shorter, easier-to-read file names that still all start with the same so they sort together

test/ads.html
```diff
@@ -1,33 +0,0 @@
-<!DOCTYPE html>
```
incompl: This file was way old and broken. I don't think it's worth fixing / maintaining

test/events.backup.js
```diff
@@ -1,3 +1,5 @@
+import '../example/app.js';
```
incompl: I didn't really fix this file, but I think this is one thing it'd need. The jury is out on whether we fix or rewrite this file.

test/events.html
```diff
@@ -1,47 +0,0 @@
-<!DOCTYPE html>
```
incompl: Also old, broken, and I don't think worth fixing

test/test.redispatch.js
```diff
@@ -0,0 +1,89 @@
...
+    // Redispatch helper for tests
+    this.redispatch = function(type) {
+      const event = {type};
+
+      eventBuffer = [];
+      redispatch.call(this.player, event);
+
+      if (eventBuffer.length === 1) {
+        return eventBuffer[0].type;
+      } else if (event.cancelBubble) {
```
ldayananda: Is this possible with how event wraps type?

test/test.redispatch.js
```diff
@@ -0,0 +1,89 @@
...
+
+      ads: {
+        state: 'content-set',
+
+        snapshot: {
+          ended: false,
+          currentSrc: 'my vid'
+        },
+
+        videoElementRecycled() {
```
ldayananda: would be really, really nice to have another test for the `return true` case at the same time.

test/test.redispatch.js
incompl: Redispatch can add cancelBubble to the mocked event, even though it's not a real event.

I added a new unit test to this PR to show that it works.

test/test.redispatch.js
incompl: Added a test

ldayananda:
LGTM

PR #245, opened 2017-03-03
Fix ended event firing

incompl:


incompl:
This code has been refactored to the point where this PR no longer has bearing on reality.

PR #250, opened 2017-04-10
Fixing ad cancellation by cues

ldayananda:
Currently, there is no way to cancel an ad as the cue gets processed before the cancellation gets a chance to happen. This moves cancellation to be before processing so ad cancellation happens as expected.

Built on top of #258

### DIFF
--- a/src/cueTextTracks.js
+++ b/src/cueTextTracks.js
@@ -24,23 +24,14 @@ export function processMetadataTracks(player, processMetadataTrack) {
   };
 
   // Text tracks are available
-  if (tracks.length > 0) {
-    for (let i = 0; i < tracks.length; i++) {
-      const track = tracks[i];
-
-      setModeAndProcess(track);
-    }
-  // Wait until text tracks are added
-  // We avoid always setting the event handler in case
-  // integrations decide to handle this separately
-  // with a different handler for the same event
-  } else {
-    tracks.addEventListener('addtrack', (event) => {
-      const track = event.track;
-
-      setModeAndProcess(track);
-    });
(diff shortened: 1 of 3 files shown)


src/cueTextTracks.js
```diff
@@ -124,15 +124,15 @@ export function processAdTrack(player, cues, processCue, cancelAds) {
       return;
     }
 
+    // Optional dynamic ad cancellation
+    if (cancelAds) {
+      cancelAds(player, cueData, startTime);
+    }
+
     // Process cue as an ad cue
     processCue(player, cueData, cueId, startTime);
```
incompl: Does it still need to get processed if it's cancelled?

ldayananda: Well, we are checking for the presence of the adCancel function - the cue in question may not actually be a cancellation cue, in which case it would need to be processed.

incompl: Oh I see, thanks!

incompl:
LGTM

src/cueTextTracks.js
```diff
@@ -124,15 +124,15 @@ export function processAdTrack(player, cues, processCue, cancelAds) {
       return;
     }
 
+    // Optional dynamic ad cancellation
+    if (cancelAds) {
+      cancelAds(player, cueData, startTime);
```
incompl: Maybe this would be clearer if it was cancelAdsHandler not cancelAds

ldayananda:
Naming change made

mrocajr:
LGTM - Passed QA based on ima@2.19.2-8 prerelease testing.

mrocajr:
LGTM - Passed QA

PR #252, opened 2017-04-13
Fix nopostroll event

mimcz:
When triggering `nopostroll` event after `contentended` in `postroll?` state, it was not directly falling back to `content-resuming` state. It only works with  `adtimeout` or `adskip` events. Event was missing in state object for `postroll?` state. I think it is handy to have this event after `contentended` especially when I request postroll break at that time so I added it and updated test suite for it.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -491,6 +491,12 @@ const contribAdsPlugin = function(options) {
             player.trigger('ended');
           }, 1);
         },
+        nopostroll() {
+          this.state = 'content-resuming';
+          window.setTimeout(function() {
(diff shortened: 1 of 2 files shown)


incompl:
LGTM

incompl:
I talked about this with another contributor and we decided nopostroll's purpose is to be sent in advance to skip a postroll slot altogether, and if you're already in postroll? state, you should use adskip instead. Does that make sense, and does that help resolve your issue?

incompl:
Closing since there has been no response, let us know if you're still interested in this.

PR #253, opened 2017-04-18
Placeholder div's CSS position and top values now match the player's

alex-barstow:
### Problem
In some cases, the placeholder `div` on iPhone does not appear in the same position as the player.
### Solution
Give the placeholder the same `position` and `top` values as the player.


ldayananda:
LGTM

PR #257, opened 2017-05-09
fix: Remove old call to player.load() during snapshot restoration.

misteroneill:
This is for players which share the video element with the ad integration. This is causing problems in Chrome/Edge with Video.js 6 due to the asynchronous nature of calling player.src().


incompl:
LGTM so long as it gets thoroughly tested

sunnyCN34:
Completed the testing on chrome, safari and firefox.  LGTM

PR #259, opened 2017-05-11
Remove error throwing for live videos.

nochev:
For live videos error is thrown after ad is shown. I found [here](https://github.com/videojs/videojs-contrib-ads/blob/v3.3.13/src/videojs.ads.js#L448) that in previous version `false` was returned instead of throwing an error.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -183,8 +183,7 @@ const contribAdsPlugin = function(options) {
     // AdBlocker is intercepting doesn't update currentSrc.
     videoElementRecycled() {
       if (!this.snapshot) {
-        throw new Error(
-          'You cannot use videoElementRecycled while there is no snapshot.');
+        return false;
       }
 
       const srcChanged = player.tech_.src() !== this.snapshot.src;
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -182,8 +182,13 @@ const contribAdsPlugin = function(options) {
     // We test both src and currentSrc because changing the src attribute to a URL that
     // AdBlocker is intercepting doesn't update currentSrc.
     videoElementRecycled() {
-      if (!this.snapshot) {
+      if (player.ads.shouldPlayContentBehindAd(player)) {
         return false;
+      }
+
+      if (!this.snapshot) {
+        throw new Error(
+          'You cannot use videoElementRecycled while there is no snapshot.');
       }
 
       const srcChanged = player.tech_.src() !== this.snapshot.src;


incompl:
This solution makes videoElementRecycled return false incorrectly in some cases. This is likely to result in bugs in other scenarios. One possible solution would need to make videoElementRecycled able to respond true/false accurately despite the snapshot being absent. Another option might be to make it so that videoElementRecycled is not called while there is no snapshot (which is how most of the code currently works).

nochev:
What are these cases? I see this error for live videos only. Is it possible to throw an error only when there is no snapshot but video is not live. Something like this:
```js
videoElementRecycled() {
      if (!this.snapshot) {
        if (player.ads.isLive(player)) {
          return false;
        }

        throw new Error(
          'You cannot use videoElementRecycled while there is no snapshot.');
      }
...
}
```

incompl:
What line of code is calling videoElementRecycled that is getting the error? That is the first place I'd suggest looking for a fix. That, or create a way for videoElementRecycled to work without relying on the snapshot (this is a refactor we've been making [elsewhere](https://github.com/videojs/videojs-contrib-ads/pull/251): snapshot is used for too many unrelated things and it causes problems like this one).

Your second proposed change would still make videoElementRecycled return false when there is no snapshot and when the video is live in cases where the video element would be recycled for ad playback such as on iOS.

I understand the desire to fix the bug this way. Even if there are no errors in your integration with this fix, having this videoElementReycled fail to fulfill its contract is a landmine for other integrations that use contrib-ads.

We've thought about this problem before and have not resolved it yet. Currently live ad support is provided only through [cue metadata track](https://github.com/videojs/videojs-contrib-ads#cue-metadata-tracks). Supporting standard prerolls / postrolls on live videos is a great feature that we'd love to have. It's just important that it's done in a way that doesn't risk breaking existing integrations.

nochev:
I forgot to mention that this error came after trying to upgrade from version `3.3.13` to `4.2.6`. With the same live stream I do not have any issues with previous version.  I see two places from where error is thrown: [here](https://github.com/videojs/videojs-contrib-ads/blob/master/src/redispatch.js#L86) and [here](https://github.com/videojs/videojs-contrib-ads/blob/master/src/redispatch.js#L64). As I see [here](https://github.com/videojs/videojs-contrib-ads/blob/master/src/plugin.js#L350), snapshot for live videos in OS different from IOS/ANDROID won't be ever created.  So probably the check could be same as when creating it. Something like this: 

```js
videoElementRecycled() {
      if (!this.snapshot) {
        if (player.ads.shouldPlayContentBehindAd(player)) {
          return false;
        }

        throw new Error(
          'You cannot use videoElementRecycled while there is no snapshot.');
      }
...
}
```

incompl:
Ah! I like where you're going with this. You could even check `shouldPlayContentBehindAd right off the bat:

```
  videoElementRecycled() {
    if (player.ads.shouldPlayContentBehindAd(player)) {
      return false;
    }

    if (!this.snapshot) {
      throw new Error(
        'You cannot use videoElementRecycled while there is no snapshot.');
    }
    ...
  }
```

We'll still want to get that snapshot check out of there at some point, but this looks like a reasonable fix in the meantime. If you make a PR for this I can help test it against other integrations.

nochev:
Ok, I will do it in this PR.

nochev:
Hello @incompl ,
Is there any issues with tests against other integrations ? Is this solution reasonable at this stage ?

incompl:
I tested this with a couple plugins and it looks good. Thanks for the contribution!

nochev:
Great, when you intend to create a new release including this fix ?

incompl:
Just released 4.2.7 with this fix.

nochev:
Great, thanks.

PR #260, opened 2017-05-12
Add withCredentials property to snapshot

alex-barstow:
### Description
The snapshot doesn't currently copy the `withCredentials` property from `.currentSource()`, which can prevent the playback of tokened content after a preroll. This PR will add conditional logic to `snapshot.js` that will ensure `withCredentials` is copied and restored to the source object between ads and main content.

#### Alternatively...
A better longterm solution might be to modify the snapshot to copy `player.currentSource()`, an object that itself contains a number of the source properties that right now are copied individually within the snapshot. Copying and restoring all of `player.currentSource()` would guarantee the restoration of important source properties like `withCredentials`, any DRM key systems, etc, and remove the need for more conditional logic in `snapshot.js` to handle those properties.
```js
  const snapshotObject = {
    ended: player.ended(),
    currentSrc: player.currentSrc(),
    src: player.tech_.src(),
    currentTime,
    type: player.currentType()
  };
```
would become something like...
```js
  const snapshotObject = {
    sourceObject: player.currentSource(),
    ended: player.ended(),
    src: player.tech_.src(),
    currentTime,
  };
```
Just a thought.

### DIFF
--- a/src/snapshot.js
+++ b/src/snapshot.js
@@ -48,8 +48,7 @@ export function getPlayerSnapshot(player) {
     snapshotObject.nativePoster = tech.poster;
     snapshotObject.style = tech.getAttribute('style');
 
-    // Why does `player.currentSource().withCredentials` work but not `player.tech_.currentSource_.withCredentials`?
-    if (player.currentSource().withCredentials) {
+    if (player.currentSource().withCredentials !== undefined) {
       snapshotObject.withCredentials = player.currentSource().withCredentials;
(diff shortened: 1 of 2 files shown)


incompl:
LGTM, have you been able to test this with a legit setup yet?

alex-barstow:
Going to close this and brainstorm the feasibility of refactoring the snapshot to use `player.currentSource()`.

PR #262, opened 2017-05-26
Clear registered timeouts when player is disposed.

nochev:
Sometimes when player is disposed there is an error like this:
```js
Uncaught TypeError: Cannot read property 'vdata1495796386723' of null
    at Object.hasElData (dom.js:335)
    at Object.trigger (events.js:390)
    at Player.trigger (component.js:816)
    at plugin.js:323
```
As we can see here https://github.com/videojs/videojs-contrib-ads/blob/master/src/plugin.js#L323 when window timeouts are not cleared after dispose `video.js` could not find the reference to player to trigger the event.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -641,6 +641,25 @@ const contribAdsPlugin = function(options) {
 
   ]), processEvent);
 
+  // Clear timeouts when player is disposed
+  player.on('dispose', function() {
+    if (player.ads.adTimeoutTimeout) {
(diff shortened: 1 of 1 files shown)


incompl:
Very nice, I've seen this error very rarely from the automated tests I've been working on.

LGTM, will merge after a second lgtm from another contributor.

nochev:
Great, thanks. Is it possible to make a new release including this fix?

incompl:
@nochev Released 4.2.8

PR #266, opened 2017-06-28
fix issue where captions were showing during ads on iOS

alex-barstow:
## About
These additions address a behavior in iOS Safari that can result in captions displaying during ad playback: 

When a user manually turns on native captions in the fullscreen iOS player, "the device will set an internal preference at the OS level for captions in that language. This preference survives clearing the browser cache and rebooting the phone” (from the Video.js [docs](https://github.com/videojs/video.js/wiki/Text-Tracks-Browser-Behavior#ios)). Safari evidently associates this caption preference with that specific video source and `srclang` (I haven't been able to find this in any Apple documentation, but the behavior observed while debugging this issue is pretty conclusive evidence that this is the case).

So, if a user manually turns on native captions _during ad playback_, then watches that same ad on the same device again without native captions explicitly set to ‘Off,’ Safari will switch the caption TextTrack `mode` from `"disabled"` (set by the snapshot at the beginning of ad playback) to `"showing"`.

## Solution

Fair warning: it's a bit of a hack, but possibly the only straight-forward way to have a player react to this iOS behavior. At the time the snapshot is taken and ad playback begins, add a `change` event listener to the player's `TextTrackList`. If a `change` event occurs while an ad is playing and any of the text track modes are `"showing"`, reset them to `"disabled"`. When the snapshot is restored, remove the event listener to ensure it doesn't persist into main content if there was no `change` event.

### *Limitations of this approach
Unfortunately there is no easy way to determine if a _user_ or _Safari_ has initiated a given track mode change, so in addition to re-disabling tracks that are enabled by Safari, this code will also disable any track that is enabled by a user during ad playback.

**Final Note:** I also changed the `browserNoActivityTimeout` in `karma.conf.js` from 20 seconds to 5 minutes to match the config in Video.js. I was not able to run tests or push my branch because the timeout was too short.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -175,7 +175,7 @@ const contribAdsPlugin = function(options) {
     // TODO reset state to content-set here instead of in every contentupdate case
     reset() {
       const textTrackList = player.textTracks();
-      
+
       textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);
       player.ads.disableNextSnapshotRestore = false;
       player.ads._contentEnding = false;
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -179,7 +179,7 @@ const contribAdsPlugin = function(options) {
 
         textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);
       }
-      
+
       player.ads.disableNextSnapshotRestore = false;
       player.ads._contentEnding = false;
       player.ads._contentHasEnded = false;
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -174,9 +174,12 @@ const contribAdsPlugin = function(options) {
 
     // TODO reset state to content-set here instead of in every contentupdate case
     reset() {
-      const textTrackList = player.textTracks();
+      if (player.ads.snapshot.trackChangeHandler) {
+        const textTrackList = player.textTracks();
 
-      textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);
+        textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);
+      }
+      
       player.ads.disableNextSnapshotRestore = false;
       player.ads._contentEnding = false;
       player.ads._contentHasEnded = false;
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -174,12 +174,6 @@ const contribAdsPlugin = function(options) {
 
     // TODO reset state to content-set here instead of in every contentupdate case
     reset() {
-      if (player.ads.snapshot.trackChangeHandler) {
-        const textTrackList = player.textTracks();
-
-        textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);
-      }
-
       player.ads.disableNextSnapshotRestore = false;
       player.ads._contentEnding = false;
       player.ads._contentHasEnded = false;
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -174,6 +174,9 @@ const contribAdsPlugin = function(options) {
 
     // TODO reset state to content-set here instead of in every contentupdate case
     reset() {
+      const textTrackList = player.textTracks();
+      textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);
+
       player.ads.disableNextSnapshotRestore = false;
       player.ads._contentEnding = false;
       player.ads._contentHasEnded = false;
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -175,8 +175,8 @@ const contribAdsPlugin = function(options) {
     // TODO reset state to content-set here instead of in every contentupdate case
     reset() {
       const textTrackList = player.textTracks();
+      
       textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);
-
       player.ads.disableNextSnapshotRestore = false;
       player.ads._contentEnding = false;
       player.ads._contentHasEnded = false;
--- a/src/snapshot.js
+++ b/src/snapshot.js
@@ -71,7 +71,7 @@ export function getPlayerSnapshot(player) {
 
   // iOS Safari will change caption mode to 'showing' if a user previously
   // turned captions on manually for that video source, so this will re-disable
-  // them if that occurs during ad playback
+  // them in case that occurs during ad playback
   if (videojs.browser.IS_IOS && !Array.isArray(tracks)) {
     tracks.one('change', function(event) {
       if (player.ads.state === 'ad-playback') {
--- a/src/snapshot.js
+++ b/src/snapshot.js
@@ -88,7 +88,7 @@ export function getPlayerSnapshot(player) {
     }
   };
 
-  if (videojs.browser.IS_IOS && !Array.isArray(tracks)) {
+  if (videojs.browser.IS_IOS && player.tech_.featuresNativeTextTracks && !Array.isArray(tracks)) {
     tracks.addEventListener('change', iOSTrackListChangeHandler);
   }
 
@@ -120,7 +120,7 @@ export function restorePlayerSnapshot(player, snapshotObject) {
 
   // ensures that the iOS TextTrackList 'change' listener added in getPlayerSnapshot()
   // doesn't persist into main content
-  if (videojs.browser.IS_IOS) {
+  if (videojs.browser.IS_IOS && player.tech_.featuresNativeTextTracks) {
     const tracks = player.textTracks();
 
     tracks.removeEventListener('change', snapshotObject.trackChangeHandler);
--- a/src/snapshot.js
+++ b/src/snapshot.js
@@ -70,10 +70,10 @@ export function getPlayerSnapshot(player) {
   snapshotObject.suppressedTracks = suppressedTracks;
 
   // iOS Safari will change caption mode to 'showing' if a user previously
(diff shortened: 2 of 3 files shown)


src/snapshot.js
```diff
@@ -93,6 +110,15 @@ export function restorePlayerSnapshot(player, snapshotObject) {
 
   const suppressedRemoteTracks = snapshotObject.suppressedRemoteTracks;
   const suppressedTracks = snapshotObject.suppressedTracks;
+
+  // ensures that the iOS TextTrackList 'change' listener added in getPlayerSnapshot()
+  // doesn't persist into main content
+  if (videojs.browser.IS_IOS) {
+    const tracks = player.textTracks();
+
+    tracks.off('change');
```
gkatsev: this will remove *all* change listeners, so, please store the handler somewhere so it can be passed in here.

src/snapshot.js
```diff
@@ -69,6 +69,23 @@ export function getPlayerSnapshot(player) {
   }
   snapshotObject.suppressedTracks = suppressedTracks;
 
+  // iOS Safari will change caption mode to 'showing' if a user previously
+  // turned captions on manually for that video source, so this will re-disable
+  // them in case that occurs during ad playback
+  if (videojs.browser.IS_IOS && !Array.isArray(tracks)) {
+    tracks.one('change', function(event) {
+      if (player.ads.state === 'ad-playback') {
+        const trackList = event.target.tracks_;
```
gkatsev: Probably should ask for the TextTrackList from the player again instead of using a private property.

src/snapshot.js
alex-barstow: Yep that's an oversight on my part. Whoops! Will break the handler out into a named function.

src/snapshot.js
alex-barstow: I used the `tracks_` property here because I needed an array to loop through with `forEach()`. Is there another way to access the array of TextTracks without using that private property?

src/snapshot.js
alex-barstow: I realized that by declaring the handler outside of both `getPlayerSnapshot()` and `restorePlayerSnapshot()`, it no longer has access to the `player` unless I pass it in as an argument. But since `.on()` and `.off()` expect to be passed a function reference, I'm not sure what the best practice is for passing additional parameters. What do you recommend?

src/snapshot.js
gkatsev: You can use a regular old forloop or you can delegate it to the forEach method.
```js
Array.prototype.forEach(player.textTracks(), track => track.mode === 'showing' ? track.mode = 'disabled' : '');
```

(except don't actually use the ternary there :P)

src/snapshot.js
gkatsev: Probably the easiest way is to still define the method inside of `getPlayerSnapshot()` but then keep a reference to it somewhere to be available in `restorePlayerSnapshot()`.

alex-barstow:
@gkatsev A few updates: 

- `.one()` turned out to be insufficient. There is no guarantee that the first track `change` event during ad playback will be the result of Safari enabling the track. Though the potential downside of using `.on()` is that we will forcibly disable all text tracks during ads, even if a user tries to turn on native captions multiple times. 

- Additionally, the event handler is now declared inside of `getPlayerSnapshot()`, and is passed to `restorePlayerSnapshot()` via the `snapshotObject`. Hopefully that is sufficient, although I know that the purpose of the snapshot is to store _player state_, which the event handler admittedly is not, so let me know what you think

src/snapshot.js
```diff
@@ -69,6 +69,31 @@ export function getPlayerSnapshot(player) {
   }
   snapshotObject.suppressedTracks = suppressedTracks;
 
+  /**
+   * iOS Safari will change caption mode to 'showing' if a user previously
+   * turned captions on manually for that video source, so this TextTrackList
+   * 'change' event handler will re-disable them in case that occurs during ad playback
+   */
+  const iOSTrackListChangeHandler = function() {
+    if (player.ads.state === 'ad-playback') {
```
incompl: Post-contrib-ads-5 the preferred method is `player.ads.isAdPlaying()`

incompl (changes requested):

src/snapshot.js
alex-barstow: Change made.

src/snapshot.js
```diff
@@ -69,6 +69,31 @@ export function getPlayerSnapshot(player) {
...
+   * 'change' event handler will re-disable them in case that occurs during ad playback
+   */
+  const iOSTrackListChangeHandler = function() {
+    if (player.ads.isAdPlaying()) {
+      const textTrackList = player.textTracks();
+
+      for (let i = 0; i < textTrackList.length; i++) {
+        const track = textTrackList[i];
+
+        if (track.mode === 'showing') {
```
ldayananda: I wonder if the snapshot's saved mode for this track should be updated if the browser expects a specific track to be showing after ads? Not that this is necessarily wrong.

src/snapshot.js
```diff
@@ -69,6 +69,31 @@ export function getPlayerSnapshot(player) {
...
+          track.mode = 'disabled';
+        }
+      }
+    }
+  };
+
+  if (videojs.browser.IS_IOS && !Array.isArray(tracks)) {
```
ldayananda: Does this happen if we are playing inline as well?

src/snapshot.js
```diff
@@ -93,6 +118,15 @@ export function restorePlayerSnapshot(player, snapshotObject) {
 
   const suppressedRemoteTracks = snapshotObject.suppressedRemoteTracks;
   const suppressedTracks = snapshotObject.suppressedTracks;
+
+  // ensures that the iOS TextTrackList 'change' listener added in getPlayerSnapshot()
+  // doesn't persist into main content
+  if (videojs.browser.IS_IOS) {
+    const tracks = player.textTracks();
+
+    tracks.off('change', snapshotObject.trackChangeHandler);
```
ldayananda: This should also be cleaned up on contentupdate events in the reset handler: https://github.com/videojs/videojs-contrib-ads/blob/master/src/plugin.js#L193

ldayananda (changes requested):

src/snapshot.js
alex-barstow: This only happens with native text tracks.

src/snapshot.js
alex-barstow: Not sure I follow. Shouldn't that already be handled in the restoration of the snapshot?

ldayananda: I mean in `getPlayerSnapshot`, the snapshot object is built with i. a reference to the track, ii. the original track state. If Safari chooses to make a track's mode `showing` and the original track state saved in the snapshot isn't showing, should we update the original track state saved in the snapshot to be what the browser expects or leave it as it was when the snapshot was taken?

Let me know if that was confusing =P

src/snapshot.js
ldayananda: We use native text tracks on Safari desktop as well if I remember correctly. Should desktop Safari be captured here?

alex-barstow: I'll look into that

alex-barstow: Not an issue on desktop Safari since the user has no way of enabling captions during ad playback

src/snapshot.js
alex-barstow: let me know if my change lgty

src/snapshot.js
ldayananda: I think we talked about including `featuresNativeTextTracks ` in this conditional, how do you feel about that?

src/snapshot.js
```diff
@@ -93,6 +117,15 @@ export function restorePlayerSnapshot(player, snapshotObject) {
 
   const suppressedRemoteTracks = snapshotObject.suppressedRemoteTracks;
   const suppressedTracks = snapshotObject.suppressedTracks;
+
+  // ensures that the iOS TextTrackList 'change' listener added in getPlayerSnapshot()
+  // doesn't persist into main content
+  if (videojs.browser.IS_IOS) {
```
ldayananda: I suppose you would want to specify `featuresNativeTextTracks` here too if we do in getPlayerSnapshot

ldayananda (commented):
Apart from my last few comments, this is looking very close to being ready!

src/plugin.js
```diff
@@ -174,6 +174,12 @@ const contribAdsPlugin = function(options) {
 
     // TODO reset state to content-set here instead of in every contentupdate case
     reset() {
+      if (player.ads.snapshot.trackChangeHandler) {
```
incompl: Snapshot logic should be encapsulated in snapshot.js

incompl (changes requested):

src/plugin.js
alex-barstow: This listener is also removed when `restorePlayerSnapshot` is called. I added it here so it will also be cleaned up on contentupdate events per @ldayananda's request.

incompl: Sorry but I'm going to have to disagree with Lahiru here. Not only is this not maintaining encapsulation, but having the same cleanup happen in two places makes the code difficult to reason about.

ldayananda (approved):

incompl (approved):

sunnyCN34:
The main issue of captions showing up during ad play is resolved. Hence we can release it

LGTM

PR #272, opened 2017-07-12
refactor: Better support for multiple module systems.

misteroneill:
This brings contrib-ads up to date with the latest tooling and conventions provided by the plugin generator.

- Avoids breaking changes by preserving Bower support and old `dist/` filenames.
- Adds husky instead of the deprecated ghooks project.
- Switches from the old chg thing to conventional-changelog.
- Uses Rollup instead of Browserify to produce multiple builds appropriate for different module systems.


package.json
```diff
@@ -1,138 +1,119 @@
...
-    "script": "dist/videojs.ads.min.js"
-  },
   "vjsstandard": {
     "ignore": [
       "dist",
       "docs",
-      "es5",
-      "scripts",
+      "example",
+      "test",
```
misteroneill: Ignoring this for now because lots of linter errors.

src/cueTextTracks.js
```diff
@@ -7,14 +7,16 @@
 
 import videojs from 'video.js';
 
+const cueTextTracks = {};
```
misteroneill: This file contains the only meaningful source change. The reason is that Rollup properly makes export bindings constant; so, when doing `import * as foo from '...';`, the properties of `foo` are read-only. This is how ES6 modules are supposed to work.

But contrib-ads exposes `cueTextTracks` and overrides properties of it in tests, which would throw errors with the old code because they became read-only properties.

package.json
```diff
@@ -1,138 +1,119 @@
...
-    "build:js:collapse": "bundle-collapser dist/videojs.ads.js -o dist/videojs.ads.min.js",
-    "build:js:uglify": "uglifyjs dist/videojs.ads.min.js --comments --mangle --compress --support-ie8 -o dist/videojs.ads.min.js",
-    "build:test": "babel-node scripts/build-test.js",
-    "clean": "rimraf dist test/dist es5 && mkdirp dist test/dist es5",
-    "lint": "vjsstandard src",
-    "start": "babel-node scripts/server.js",
+    "build:css": "npm-run-all build:css:sass build:css:bannerize build:css:copy",
+    "build:css:bannerize": "bannerize dist/videojs-contrib-ads.css --banner=scripts/banner.ejs",
+    "build:css:sass": "node-sass src/plugin.scss dist/videojs-contrib-ads.css --output-style=compressed --linefeed=lf",
+    "build:css:copy": "cpy dist/videojs-contrib-ads.css dist/ --rename=video.ads.css",
```
misteroneill: There are three scripts like this that aim to preserve the old filenames (`video.ads.*`) vs. the more conventional names (`videojs-contrib-ads.*`) - `dist/` will contain _both_.

It was easier to do this instead of trying to make every piece of tooling adapt to the old filenames.

package.json
```diff
@@ -1,138 +1,119 @@
 {
   "name": "videojs-contrib-ads",
   "description": "A framework that provides common functionality needed by video advertisement libraries working with video.js.",
-  "main": "es5/plugin.js",
+  "main": "dist/videojs-contrib-ads.cjs.js",
+  "module": "dist/videojs-contrib-ads.es.js",
```
misteroneill: These two changes allow support for different module systems.

incompl:
LGTM

PR #278, opened 2017-08-25
add vjs-waiting while waiting for ad to load

RevinKey:
This will add the vjs-waiting class to the player if in ads-mode but nothing is currently playing. the purpose of this is to make sure the loading spinner is animated when the player is waiting for an ad. In VPAID cases this class is not added currently and the spinner does not animate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+<a name="5.1.0"></a>
+# [5.1.0](https://github.com/videojs/videojs-contrib-ads/compare/v3.3.13...v5.1.0) (2017-09-12)
+
+### Bug Fixes
+
+* Fix mis-named dists and potentially breaking change in package.json 'main' field. ([#280](https://github.com/videojs/videojs-contrib-ads/issues/280)) ([7633161](https://github.com/videojs/videojs-contrib-ads/commit/7633161))
+* Remove old call to player.load() during snapshot restoration for players which share the video element with the ad integration. This is causing problems in Chrome/Edge with Video.js 6 due to the asynchronous nature of calling player.src(). ([#257](https://github.com/videojs/videojs-contrib-ads/issues/257)) ([afb3ccf](https://github.com/videojs/videojs-contrib-ads/commit/afb3ccf))
+
+### Chores
+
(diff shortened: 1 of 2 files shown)


src/plugin.js
```diff
@@ -300,18 +300,16 @@ const contribAdsPlugin = function(options) {
...
-
-             // Waiting to learn about preroll
-             player.ads.state === 'preroll?' ||
-
-             // A linear ad is active
-             player.ads.state === 'ad-playback' ||
-
-             // Content is not playing again yet
-             player.ads.state === 'content-resuming';
+      if (player.ads.state === 'ads-ready?' || // Waiting to learn about preroll
```
gkatsev: Tests are failing because this is against our style guide. Comments should go above the code they're commenting. Can you revert this change?

src/plugin.js
```diff
@@ -300,18 +300,16 @@ const contribAdsPlugin = function(options) {
...
+          player.ads.state === 'preroll?' || // A linear ad is active
+          player.ads.state === 'ad-playback' || // Content is not playing again yet
+          player.ads.state === 'content-resuming') {
+        player.addClass('vjs-waiting');
+        return true;
```
gkatsev: We should instead store the value of the boolean expression and check it's value for whether to add or remove the class and then return it rather than return booleans directly.

src/plugin.js
```diff
@@ -323,7 +321,12 @@ const contribAdsPlugin = function(options) {
     // This relies on startLinearAdMode and endLinearAdMode because that is the
     // most authoritative way of determinining if an ad is playing.
     isAdPlaying() {
-      return this._inLinearAdMode;
+      if (this._inLinearAdMode) {
```
gkatsev: It would be better to write this as
```js
if (this._inLinearAdMode) {
  player.removeClass('vjs-waiting');
}

return this._inLinearAdMode;
```

incompl:
This check should not be in `isInAdMode`. This means a loading spinner could be inadvertently added when ad mode is checked in other cases. I recommend looking at the places where `isInAdMode` is invoked to look for a place that is more reliable.

PR #279, opened 2017-09-01
feat: add a cancelContentTimeout option

BrandonOCasey:
This is a workaround for a problem that we are seeing which is going to be rather hard to solve. 

The basic gist comes from this line of code: 
https://github.com/videojs/videojs-contrib-ads/blob/master/src/cancelContentPlay.js#L69

A play promise is in flight as this code runs so the player is not technically playing (although it will be). So we think that we need to pause the video for contrib-ads loading. This cancels the promise and the state of play is lost. This causes the play button to show back up during a preroll even though the user just clicked it.

### DIFF
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -58,6 +58,9 @@ export default function cancelContentPlay(player) {
 
   }
 
+  if (!player.ads.doCancelPlayTimeout) {
+    return;
+  }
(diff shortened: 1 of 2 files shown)


misteroneill:
I wonder if, instead of adding a new option, it would be better to use the existing `stitchedAds` option here.

misteroneill:
As discussed elsewhere, I think we're going to take a different approach here.

PR #280, opened 2017-09-01
fix: Fix mis-named dists and potentially breaking change in package.json 'main' field.

misteroneill:
In #272, I mis-translated the names of the `dist/` files to `video.ads.*` instead of `videojs.ads.*`. Also, the change to `main` in `package.json` could potentially be considered breaking because it _may_ affect certain Browserify-based builds.


incompl:
LGTM

PR #283, opened 2017-09-08
Fix master.

misteroneill:
Looks like I pushed to the wrong remote. Also, noticed that there was an error in the package.json migration due to this project being a bit different and needing to continue supporting Bower!


package.json
```diff
@@ -37,6 +37,7 @@
     "test": "karma start test/karma.conf.js",
     "preversion": "npm test",
     "version": "node scripts/version.js",
+    "postversion": "node scripts/postversion.js",
```
misteroneill: This is the only actual change (besides removing `dist`).

incompl:
LGTM

PR #284, opened 2017-09-08
More complete support for stitched ad scenarios.

misteroneill:
- Use player timeout methods to avoid even the slight risk of memory leaks (and obviating the need to clean them up manually on dispose).
- Do not cancel content play when using stitched ads.
- Do not use snapshots when playing stitched ads.
- Avoid loading states when playing stitched ads.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -117,6 +117,12 @@ const contribAdsPlugin = function(options) {
       return;
     }
 
+    // With stitched ads, you may see a durationchange at the end of ad playback
+    // when switching back to content mode. We shouldn't need to call play().
+    if (player.ads.stitchedAds()) {
+      return;
+    }
+
     player.play();
   });
 
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -231,6 +231,10 @@ const contribAdsPlugin = function(options) {
       }
     },
 
+    // Get/set whether the content source includes stitched ads. This changes
+    // the behavior of contrib-ads to avoid behaviors that are required for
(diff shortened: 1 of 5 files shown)


incompl:
Looks awesome. Is player.setTimeout new in some version of videojs or did we just miss it?

misteroneill:
It's been in Video.js for a long time (supported by the minimum dependent version here). It's not a well-known feature, though.

Also has `player.requestAnimationFrame` if that's ever useful!

src/plugin.js
```diff
@@ -273,12 +275,18 @@ const contribAdsPlugin = function(options) {
     // Return true if content playback should mute and continue during ad breaks.
     // This is only done during live streams on platforms where it's supported.
     // This improves speed and accuracy when returning from an ad break.
-    shouldPlayContentBehindAd(somePlayer) {
+    shouldPlayContentBehindAd(somePlayer = player) {
```
misteroneill: Might as well default to the in-scope player.

incompl:
Is this still a WIP / still incoming?

misteroneill:
@incompl So far, I haven't seen anything to indicate this work is insufficient for our stitched ads use-case; so, I think it's fair to say it's no longer WIP. 👍

incompl:
Great in that case I've reviewed and this LGTM.

misteroneill:
Re-implementing this against contrib-ads 6 would be more than a rebase. It would be best to take an approach that reflects the recent architectural changes. I won't have time in the near-term to do that, though. Closing, for now.

PR #292, opened 2017-10-27
fix: TypeError - player.ads.snapshot.trackChangeHandler

alex-barstow:
This will fix `VIDEOJS: ERROR: TypeError: Cannot read property 'trackChangeHandler' of null` errors on `contentupdate` events when the snapshot is not used.

### DIFF
--- a/src/snapshot.js
+++ b/src/snapshot.js
@@ -9,7 +9,7 @@ import videojs from 'video.js';
 
 function initialize(player) {
   player.on('contentupdate', function() {
-    if (player.ads.snapshot.trackChangeHandler) {
+    if (player.ads.snapshot && player.ads.snapshot.trackChangeHandler) {
       const textTrackList = player.textTracks();
 
       textTrackList.removeEventListener('change', player.ads.snapshot.trackChangeHandler);


src/snapshot.js
```diff
@@ -9,7 +9,7 @@ import videojs from 'video.js';
 
 function initialize(player) {
   player.on('contentupdate', function() {
-    if (player.ads.snapshot.trackChangeHandler) {
+    if (player.ads.snapshot && player.ads.snapshot.trackChangeHandler) {
```
incompl: I'm concerned about the underlying cause: how could there be a snapshot object before snapshot is initialized?

incompl: After looking at this in more detail, I believe the underlying issue is that `player.ads.snapshot` is not the right place to store `trackChangeHandler`. From a domain perspective, this handler isn't part of the snapshot state. From a more concrete perspective, snapshot objects can come and go as a matter of course, so you can't rely on them being there at any time. I suggest storing trackChangeHandler in a module-level variable instead.

alex-barstow: @incompl @ldayananda  **1)** I agree that the snapshot is not the most logical place to store the handler. **2)** After looking closer, there seems to be a fundamental flaw in the logic inside of `initialize()`. Every time there is a contentupdate, `player.ads.snapshot` is set to `null` in [reset()](https://github.com/videojs/videojs-contrib-ads/blob/master/src/plugin.js#L196), so this conditional will always be falsy and the `trackChangeHandler` event listener never gets removed.

ldayananda: 2) Is a great observation and I think will be solved by storing the `trackChangeHandler` on the plugin rather than the snapshot object as @incompl suggested

ldayananda:
Thanks for the change! This bug has been addressed by #293

PR #293, opened 2017-10-28
fix: snapshot.trackChangeHandler is undefined

ldayananda:
Changes include:

- now calls `snapshot. initializeSnapshot` when setting up the plugin
- now adds and removes the event handler to all platforms when creating/restoring a snapshot
- unit tests

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -740,34 +740,44 @@ const contribAdsPlugin = function(options) {
 
   };
 
+  // A utility method for textTrackChangeHandler to define the conditions
+  // when text tracks should be disabled.
+  // Currently this includes:
+  //  - on iOS with native text tracks, during an ad playing
+  const shouldDisableTracks = function() {
+    // If the platform matches iOS with native text tracks
+    // and this occurs during ad playback, we should disable tracks again.
+    // If shouldPlayContentBehindAd, no special handling is needed.
+    return !player.ads.shouldPlayContentBehindAd(player) &&
+            player.ads.isAdPlaying() &&
+            player.tech_.featuresNativeTextTracks &&
+            videojs.browser.IS_IOS &&
+            // older versions of video.js did not use an emulated textTrackList
+            !Array.isArray(player.textTracks());
+  };
+
   /**
    * iOS Safari will change caption mode to 'showing' if a user previously
    * turned captions on manually for that video source, so this TextTrackList
    * 'change' event handler will re-disable them in case that occurs during ad playback
    */
-  const iOSTrackListChangeHandler = function() {
+  const textTrackChangeHandler = function() {
     const textTrackList = player.textTracks();
 
-    // Exit early if the platform does not match or this occurs during content
-    // If shouldPlayContentBehindAd, no special handling is needed
-    if (player.ads.shouldPlayContentBehindAd(player) || !player.ads.isAdPlaying() ||
-      !player.tech_.featuresNativeTextTracks || !videojs.browser.IS_IOS ||
-      Array.isArray(textTrackList)) {
-      return;
-    }
+    if (shouldDisableTracks()) {
+      // We must double check all tracks
+      for (let i = 0; i < textTrackList.length; i++) {
+        const track = textTrackList[i];
 
-    for (let i = 0; i < textTrackList.length; i++) {
-      const track = textTrackList[i];
-
-      if (track.mode === 'showing') {
-        track.mode = 'disabled';
+        if (track.mode === 'showing') {
+          track.mode = 'disabled';
+        }
       }
     }
   };
 
-  player.ads._trackChangeDuringAdsHandler = iOSTrackListChangeHandler;
   // Add the listener to the text track list
-  player.textTracks().addEventListener('change', player.ads._trackChangeDuringAdsHandler);
+  player.textTracks().addEventListener('change', textTrackChangeHandler);
 
   // Register our handler for the events that the state machine will process
   player.on(VIDEO_EVENTS.concat([
@@ -810,7 +820,7 @@ const contribAdsPlugin = function(options) {
       player.clearTimeout(player.ads.tryToResumeTimeout_);
     }
 
-    player.textTracks().removeEventListener('change', player.ads._trackChangeDuringAdsHandler);
+    player.textTracks().removeEventListener('change', textTrackChangeHandler);
   });
 
   // If we're autoplaying, the state machine will immidiately process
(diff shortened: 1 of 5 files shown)


ldayananda:
Talked to @incompl and decided to take this event further and remove it from the snapshot feature entirely.

src/plugin.js
```diff
@@ -336,6 +336,9 @@ const contribAdsPlugin = function(options) {
   // Start sending contentupdate events for this player
   initializeContentupdate(player);
 
+  // Prepare the snapshot feature for use during ad playback
+  snapshot.initializeSnapshot(player);
```
incompl: Might as well get rid of this, at this rate we may never need it.

src/plugin.js
```diff
@@ -740,6 +743,35 @@ const contribAdsPlugin = function(options) {
...
+   * iOS Safari will change caption mode to 'showing' if a user previously
+   * turned captions on manually for that video source, so this TextTrackList
+   * 'change' event handler will re-disable them in case that occurs during ad playback
+   */
+  const iOSTrackListChangeHandler = function() {
+    const textTrackList = player.textTracks();
+
+    // Exit early if the platform does not match or this occurs during content
+    // If shouldPlayContentBehindAd, no special handling is needed
+    if (player.ads.shouldPlayContentBehindAd(player) || !player.ads.isAdPlaying() ||
```
incompl: I think it'd be more readable / extensible if we reversed this and did
```
if (!player.ads.shouldPlayContentBehindAd(player) && player.ads.isAdPlaying() ...) {
  for (...) {...}
}
```

src/plugin.js
```diff
@@ -740,6 +743,35 @@ const contribAdsPlugin = function(options) {
...
+    for (let i = 0; i < textTrackList.length; i++) {
+      const track = textTrackList[i];
+
+      if (track.mode === 'showing') {
+        track.mode = 'disabled';
+      }
+    }
+  };
+
+  player.ads._trackChangeDuringAdsHandler = iOSTrackListChangeHandler;
```
incompl: We don't need to store this on the ads object anymore.

src/plugin.js
```diff
@@ -780,6 +812,10 @@ const contribAdsPlugin = function(options) {
     if (player.ads.tryToResumeTimeout_) {
       player.clearTimeout(player.ads.tryToResumeTimeout_);
     }
+
+    if (player.ads._trackChangeDuringAdsHandler) {
```
incompl: This conditional isn't necessary, it's no problem to call removeEventListener even if you don't know if it's there.

src/plugin.js
```diff
@@ -740,6 +743,35 @@ const contribAdsPlugin = function(options) {
 
   };
 
+  /**
+   * iOS Safari will change caption mode to 'showing' if a user previously
+   * turned captions on manually for that video source, so this TextTrackList
+   * 'change' event handler will re-disable them in case that occurs during ad playback
+   */
+  const iOSTrackListChangeHandler = function() {
```
incompl: At this point the name is a misnomer, we could call it something like `textTrackChangeHandler`

test/test.ads.js
```diff
@@ -1169,3 +1169,39 @@ QUnit.test('adserror ends linear ad mode ', function(assert) {
...
+      // Start the ad, confirm the track is disabled
+      this.player.ads.startLinearAdMode();
+      assert.equal(tracks[0].mode, 'disabled', 'Snapshot sets tracks to disabled');
+
+      // Force the mode to showing
+      tracks[0].mode = 'showing';
+
+    }.bind(this));
+
+    // The mode should go back to disabled when the change happens
```
incompl: Since this issue is subtle, it could be helpful to be more specific about what the change is.

src/plugin.js
ldayananda: Don't we need it to remove the listener on `dispose` still?

src/plugin.js
incompl: I believe it could mislead a reader into thinking it is only called on iOS, which is not the case.

src/plugin.js
incompl: We do, and we can remove `iOSTrackListChangeHandler` rather than a reference stored on the `ads` object. It's the same value.

ldayananda: Ah, yes `iOSTrackListChangeHandler` is in scope here, my mistake!

src/plugin.js
ldayananda: We could rename it to the more general `textTrackChangeInAdsHandler`, what do you think?

incompl: The handler is always present even if it's not in ad mode, and the check for ads is within the handler itself. `textTrackChangeHandler` is still the most accurate I can think of. More logic could be added to the handler and the name would remain accurate.

ldayananda:
Feedback incorporated :)

src/plugin.js
```diff
@@ -740,6 +740,45 @@ const contribAdsPlugin = function(options) {
...
 
+  // A utility method for textTrackChangeHandler to define the conditions
+  // when text tracks should be disabled.
+  // Currently this includes:
+  //  - on iOS with native text tracks, during an ad playing
+  const shouldDisableTracks = function() {
+    // If the platform matches iOS with native text tracks
+    // and this occurs during ad playback, we should disable tracks again.
+    // If shouldPlayContentBehindAd, no special handling is needed.
+    return !player.ads.shouldPlayContentBehindAd(player) &&
```
ldayananda: Would appreciate a close look here to make sure I did the negation correctly!

incompl: Looks right to me. Not live & ad playing & native text tracks & ios & not older vjs native text tracks

test/test.snapshot.js
```diff
@@ -437,6 +440,10 @@ QUnit.test('Snapshot and text tracks', function(assert) {
   assert.equal(this.player.textTracks()[0].language, 'es');
   assert.equal(this.player.textTracks()[0].mode, 'disabled');
 
+  // Double check that the track remains disabled
+  assert.equal(this.player.remoteTextTracks()[0].mode, 'disabled');
+  assert.equal(this.player.textTracks()[0].mode, 'disabled');
```
incompl: It looks like this assertion happens twice, previously on line 441. What could have changed it in that time?

incompl:
LGTM overall, just minor comments now

test/test.snapshot.js
ldayananda: The original reason for the textTrackChangeHandler was to mitigate user agents changing the track mode during ad playback, I added this here to make sure that doesn't happen for browsers other than iOS but looks like I forgot to add a `clock.tick()`. Will fix that now!

Issue #291, opened 2017-10-26
Player.play() doesn't always cause transition from ads-ready to preroll?


shawnbuso:
I'm the owner of the videojs-ima plugin. I'm doing a small re-write to get it up to date with the latest contrib-ads plugin. I'm running into an issue with my playlist sample - every other time I call play(), the content just plays instead of causing a transition from ads-ready to preroll?. As a result I only see ads for every other video.

Steps to repro:
1. Load my [sample](https://shawnbuso.github.io/videojs-ima/examples/playlist_simple/).
2. Open the dev console.
3. Click the little Android dude in the video 1 thumbnail (don't click the video to play it).
4. After the pre-roll, click the doubleclick logo in the video 2 thumbnail.

In my testing I see the following:
Page loads
I click on the Android thumbnail to request ads for that video
Ads plugin transitions content-set -> ads-ready
After triggering ads-ready, the implementation calls player.play()
Ads plugin transitions ads-ready -> preroll? -> ad-playback -> content-resuming -> content-playback
I click on the doubleclick thumbnail to request new ads for that video
Ads plugin transitions content-playback -> content-set -> ads-ready
After triggering ads-ready, the implementation calls player.play()
Ads plugin does not transition to preroll?, the content just plays

If you repeat this process you continue to see ads for every other video, no matter which video you start with (e.g. if you click the doubleclick ad first instead of the Android ad, then that will always have ads and the android ad will not). Any tips on what may be going wrong here?

incompl:
Howdy, I'm happy to help figure this out. What are the versions of contrib-ads you're migration from and to? That'll help me narrow down the cause.

shawnbuso:
Awesome, thanks so much! We're upgrading from VJS 5.3.0 and contrib-ads 5.0.3 to VJS 5.20.4 and contrib-ads 5.1.0.

I also just noticed we're getting an error now - when I switch videos, I'm seeing "VIDEOJS: ERROR: TypeError: Cannot read property 'trackChangeHandler' of null". Oddly enough I still see ads for every other video, even though I'm seeing this error every time I switch.

marguinbc:
[...]

PR #294, opened 2017-11-02
Added player.ads._cancelledPlay = false to reset

mimse:
This is a PR for #291


incompl:
I tested this out with a couple plugins and looked at the cancelContentPlay more closely and found that this change is 💯

PR #295, opened 2017-11-02
add vjs-waiting while waiting for ad to load

RevinKey:
In VPAID cases the loading spinner does not animate due to vjs-waiting not being added to the player while waiting fro ads to load. reopen of #278

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,24 +1,30 @@
+<a name="5.1.5"></a>
+## [5.1.5](https://github.com/videojs/videojs-contrib-ads/compare/v5.0.4-0...v5.1.5) (2017-11-21)
+
+### Bug Fixes
+
+* Remove the placeholder div logic from cancelContentPlay() ([#296](https://github.com/videojs/videojs-contrib-ads/pull/296))
+
 <a name="5.1.4"></a>
 ## [5.1.4](https://github.com/videojs/videojs-contrib-ads/compare/v5.0.4-0...v5.1.4) (2017-11-17)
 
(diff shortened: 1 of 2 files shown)


ldayananda:
@RevinKey It looks like you have a small linting issue that needs to be fixed for the CI tests to run. Do you mind doing that?

RevinKey:
@ldayananda all set it passed CI tests now.

src/plugin.js
```diff
@@ -382,6 +386,8 @@ const contribAdsPlugin = function(options) {
         if (player.ads.nopreroll_) {
           // This will start the ads manager in case there are later ads
           player.trigger('readyforpreroll');
+          // add class for spinner animation while in ?preroll state
+          player.addClass('vjs-waiting');
```
incompl: If there is no preroll, is it necessary to add vjs-waiting?

incompl: Ping @RevinKey

incompl:
I made a diagram of where the class is added and removed in your PR and found that it can be simplified.

unnamed

Here is a simplified flow that I recommend based on this:

* Add the class when leaving content-set
* Add the class when entering `postroll?`
* Remove the class on entering `content-playback`
* Remove the class on entering `ad-playback`

incompl:
LGTM

sunnyCN34:
IMA: @brightcove/videojs-ima3@2.20.4-0
@brightcove/videojs-freewheel@2.8.6-1

Issue found:

Animated spinner still visible in background on iPads (IOS 11), when page is loaded and BPB is not pressed.

PR #296, opened 2017-11-03
Remove the placeholder div logic from cancelContentPlay()

alex-barstow:
The time has come to remove the placeholder div (aka The Black Div of Death) from `cancelContentPlay()`. Content flash before preroll playback in the iOS native player (the original reason for its creation) is evidently less of an issue in later versions of iOS, and this workaround has caused more problems than it has solved. The decision to remove it comes sanctioned by @incompl and @ldayananda.

### DIFF
--- a/src/cancelContentPlay.js
+++ b/src/cancelContentPlay.js
@@ -6,9 +6,6 @@ then signalling that we should play after the ad is done.
 */
 
 import window from 'global/window';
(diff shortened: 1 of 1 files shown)


incompl:
LGTM

sunnyCN34:
QA- passed Black screen does not appear now in native player of iOS 11 iPhone.

PR #301, opened 2017-11-14
fix: Captions showing during ads on desktop Safari 11

alex-barstow:
Fix for [BC-39385](https://jira.brightcove.com/browse/BC-39385).

## Description
We had previously observed and addressed [a problem on iOS](https://github.com/videojs/videojs-contrib-ads/pull/266) where Safari will sometimes forcibly enable text tracks by changing their mode to `showing`. At the time, that behavior only occurred on mobile, however, as of Safari 11 we now occasionally see the same thing on desktop.

## Solution
Change the scope of the existing solution for iOS to include desktop Safari 11.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -32,7 +32,6 @@ const removeNativePoster = function(player) {
   }
 };
 
-
 const DESKTOP_SAFARI_VERSION = (function() {
   const USER_AGENT = window.navigator && window.navigator.userAgent || '';
   const match = USER_AGENT.match(/Version\/(\d+)/);
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -752,7 +752,7 @@ const contribAdsPlugin = function(options) {
     return !player.ads.shouldPlayContentBehindAd(player) &&
             player.ads.isAdPlaying() &&
             player.tech_.featuresNativeTextTracks &&
-            (videojs.browser.IS_IOS || DESKTOP_SAFARI_VERSION === 11) &&
+            (videojs.browser.IS_IOS || DESKTOP_SAFARI_VERSION >= 11) &&
             // older versions of video.js did not use an emulated textTrackList
             !Array.isArray(player.textTracks());
   };
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -32,6 +32,17 @@ const removeNativePoster = function(player) {
   }
 };
 
+
+const DESKTOP_SAFARI_VERSION = (function() {
+  const USER_AGENT = window.navigator && window.navigator.userAgent || '';
+  const match = USER_AGENT.match(/Version\/(\d+)/);
+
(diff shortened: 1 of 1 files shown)


src/plugin.js
```diff
@@ -742,7 +752,7 @@ const contribAdsPlugin = function(options) {
     return !player.ads.shouldPlayContentBehindAd(player) &&
             player.ads.isAdPlaying() &&
             player.tech_.featuresNativeTextTracks &&
-            videojs.browser.IS_IOS &&
+            (videojs.browser.IS_IOS || DESKTOP_SAFARI_VERSION === 11) &&
```
incompl: How about `DESKTOP_SAFARI_VERSION >= 11` instead?

alex-barstow: Yeah, that is a safer bet. I wrote it this way since we technically don't know if this fix will be needed in future versions of Safari, but including Safari 12+ does have the potential to save us some headaches down the road, and doesn't have a downside I can think of. I'll make the change.

incompl:
LGTM

src/plugin.js
```diff
@@ -32,6 +32,16 @@ const removeNativePoster = function(player) {
   }
 };
 
+const DESKTOP_SAFARI_VERSION = (function() {
```
gkatsev: if/when this gets added to Video.js, we'd probably want to update this to use that first and then fallback to this code.

alex-barstow: @gkatsev Considering what we discussed yesterday re: the problem with browser version checks returning `null` on the wrong platform, do you think we should keep this as is, or get rid of `return null` here?

gkatsev: Also, we are no longer adding DESKTOP_SAFARI_VERSION to video.js right now but maybe it would still be good to be consistent with the other ones? And then making sure to also check `IS_SAFARI`?

alex-barstow: `IS_SAFARI` will still return true on iOS, so I believe we'd need to add a `IS_DESKTOP_SAFARI` check to avoid the same `null` issue.

gkatsev: I guess you could just do `DESKTOP_SAFARI_VERSION && DESKTOP_SAFARI_VERSION >= 11`.

Or if the rest of the team don't mind the loss of consistency update this snippet not to require a null check.

alex-barstow: Actually, technically the `null` issue only poses a problem if we are checking if the browser version is `less than` a given number since null gets coerced into a 0. Since we're checking `DESKTOP_SAFARI_VERSION >= 11` here maybe we don't need to worry about doing a platform check first?

alex-barstow:
We started QA on this but haven't been able to reproduce the original problem. No sense proceeding with a fix for something we can no longer reproduce. If the issue comes up again, we can re-open this and proceed with testing.

PR #302, opened 2017-11-15
Make example init correctly

incompl:


ldayananda:
Although the tests failed so you'll want to fix that before merging

Issue #305, opened 2017-11-30
Setting source on adend leads to weirdness


damongant:
We use video.js for streaming. Depending on browser support, we pick HLS or MP4 (not DASH, our streaming server serves a [normal looking MP4 file](https://1-edge5-us-east.picarto.tv/mp4/restreamchecker.mp4?token=public&con=1512032620281&ticket=0&type=0&scope=0)) Because we keep only about the last 10 seconds of the streaming buffer on the server, playing a 15 second ad causes the following scenario:

* video.js loads the MP4 container with the stream and the first few frames
* contrib-ads pauses the stream, prepares and plays a 15 second ad
* contrib-ads unpauses the stream, the frames loaded initially play. The streaming server refuses to serve the frames that would've been loaded during the ad because it does not have them anymore (and also we want low latency) and throws a media error after a few seconds.
* Our error handler kicks in and restarts the stream after 10 seconds.

We attempted to fix this by setting the source anew (with a timestamp in the URL to force a reload, what we're already doing whenever playback is resumed for MP4 streams) in an `adend` handler.

This lead to a weird race condition with up to 3 play events being fired, the player eventually recovers after 10-15 seconds, but this takes too long. We fixed this for now by using setTimeout with 0 seconds to push the call to whatever contrib-ads does after triggering the `adend` event.

incompl:
The most success we've had providing a good experience for ads in live streams is to play the stream muted in the background while the ad plays in a container. The `shouldPlayContentBehindAd` check manages this: it returns true if the duration is Infinity (live stream) or if it's on a platform that doesn't support multiple video elements. This also means you can respond to ID3 metadata in the stream during ads if needed. Is this an option in your case?

damongant:
I will test this, but I'd be surprised if it works as expected. Because we use a normal MP4 container for livestreaming (due to compatibility issues with some devices), videojs does not detect it as a livestream - just as a very very long video. I've previously tried to override the prototype of the function that returns the video duration to always return `Infinity` and while that fixes automatic live display and the likes, it also breaks MP4 playback after a short while.

incompl:
Ahh interesting. I think what we need for you is a way to explicitly opt-in to the live behavior. I made a branch that adds a `live` option to do just that. Would you be willing to give it a shot?

https://github.com/videojs/videojs-contrib-ads/tree/live-option

damongant:
I'll give that a go on Monday, thanks for putting the work in for us!

[...]

PR #306, opened 2017-12-01
Live option

incompl:
See #305


ldayananda:
@incompl are you looking for review on this?

incompl:
No, the issue was resolved another way.

PR #308, opened 2017-12-19
Fix caption persistence

alex-barstow:
Proposed fix for [BC-38559](https://jira.brightcove.com/browse/BC-38559) and [BC-39463](https://jira.brightcove.com/browse/BC-39463).

## Description
There are 3 issues this PR addresses:

1. Text track modes are not correctly restored after ads in general. This is the result of the fact that the snapshot saves and restores both `player.textTracks()` and `player.remoteTextTracks()` separately. `player.textTracks()` contains the TextTrackList of all text tracks, so it isn't necessary to also get a snapshot of `remoteTextTracks()`. And since they both store references to the same remote text tracks, there's a bug in [this logic](https://github.com/videojs/videojs-contrib-ads/blob/master/src/snapshot.js#L49-L69) where the remote text track modes are all disabled before the second loop, so the snapshot saves and restores them as such.

2. There is also a separate (but related) issue where `restorePlayerSnapshot()` [doesn't restore text track modes](https://github.com/videojs/videojs-contrib-ads/blob/master/src/snapshot.js#L225-L227) after a postroll ad. This means that caption preference will not persist through multiple videos in a playlist.

3. If there isn't a postroll, caption preference still will not persist through videos in a playlist because the snapshot taken in the `postroll?` state disables all track modes and they are never restored.

## Specific changes
1. Remove `postroll?` snapshot
2. Modify conditional in `restorePlayerSnapshot()` to restore text track modes after postroll ads
3. Only save and restore `player.textTracks()`, no need to snapshot `remoteTextTracks()`
4. Make necessary modifications to tests (also addressed an issue I discovered where [this test](https://github.com/videojs/videojs-contrib-ads/blob/master/test/test.snapshot.js#L297) seems to never fail because `tracks` is always an array with length 0.)

### DIFF
--- a/src/snapshot.js
+++ b/src/snapshot.js
@@ -29,9 +29,7 @@ export function getPlayerSnapshot(player) {
   }
 
   const tech = player.$('.vjs-tech');
-  const remoteTracks = player.remoteTextTracks ? player.remoteTextTracks() : [];
   const tracks = player.textTracks ? player.textTracks() : [];
-  const suppressedRemoteTracks = [];
   const suppressedTracks = [];
   const snapshotObject = {
     ended: player.ended(),
@@ -45,17 +43,6 @@ export function getPlayerSnapshot(player) {
     snapshotObject.nativePoster = tech.poster;
     snapshotObject.style = tech.getAttribute('style');
   }
-
-  for (let i = 0; i < remoteTracks.length; i++) {
-    const track = remoteTracks[i];
-
-    suppressedRemoteTracks.push({
-      track,
-      mode: track.mode
-    });
-    track.mode = 'disabled';
-  }
-  snapshotObject.suppressedRemoteTracks = suppressedRemoteTracks;
 
   for (let i = 0; i < tracks.length; i++) {
     const track = tracks[i];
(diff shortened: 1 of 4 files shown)


src/snapshot.js
```diff
@@ -222,11 +203,14 @@ export function restorePlayerSnapshot(player, snapshotObject) {
...
     player.ads.tryToResumeTimeout_ = player.setTimeout(tryToResume, 2000);
-  } else if (!player.ended() || !snapshotObject.ended) {
+  } else {
     // if we didn't change the src, just restore the tracks
     restoreTracks();
-    // the src didn't change and this wasn't a postroll
-    // just resume playback at the current time.
-    player.play();
+
+    if (!player.ended() || !snapshotObject.ended) {
```
incompl: Do you know why it checks both `player.ended()` and `snapshotObject.ended`? It seems like one should be enough, so a comment would help at least.

gkatsev: Looks like it comes from here: https://github.com/videojs/videojs-contrib-ads/commit/a11cbce454c56cbf214a37908ba82bad8fe12a36, it seems like it's to support postrolls?

alex-barstow: I just left this conditional as it already was, but with that said I can't think of a case where `!player.ended()` and `!snapshotObject.ended` would not have the same value, but I may be missing something.

gkatsev: server-side ads could potentially have a snapshot of ended but a player not ended at least at the time of the initial writing. I think we don't do snapshots for that kind of video anymore?

incompl: I guess we could either leave it as is or launch a detailed investigation and I'm fine with both hah. I just want to make sure we build more clarity around these sorts of details over time.

alex-barstow: I'm all for a detailed investigation, but I don't think there's any reason it should hold up these bugs getting fixed.

incompl: I'm going to investigate right now

alex-barstow: Following investigation, we are going to just check for `!player.ended()` here since if we reach this line the video element was not recycled.

incompl:
LGTM, thanks again for helping improve the code.

mrocajr:
Passed QA

PR #309, opened 2017-12-20
CSS to make sure spinner is animated during waiting for ads

RevinKey:
Add CSS for spinning animation to target `.vjs-ad-playing.vjs-ad-loading` so that certain ads and waiting states will animate the spinner component.

### DIFF
--- a/src/plugin.scss
+++ b/src/plugin.scss
@@ -26,8 +26,8 @@ Styles for ad integrations.
     display: none;
 }
 
-/* Following is copied from https://github.com/videojs/video.js/blob/master/src/css/components/_loading.scss in order to
- this is needed to animate the spinner when ad specifically VPAIDs are loading.
+/* Following is copied from https://github.com/videojs/video.js/blob/master/src/css/components/_loading.scss 
+ this is needed to animate the spinner when ads, specifically VPAIDs, are loading.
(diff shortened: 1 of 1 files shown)


src/plugin.scss
```diff
@@ -25,3 +25,37 @@ Styles for ad integrations.
 .vjs-ad-playing .vjs-audio-button {
     display: none;
 }
+// only animate when showing because it can be processor heavy
+
+.vjs-ad-playing.vjs-ad-loading .vjs-loading-spinner {
```
incompl: I think we can skip this block because it won't be shown by the ad plugin.

src/plugin.scss
```diff
@@ -25,3 +25,20 @@ Styles for ad integrations.
 .vjs-ad-playing .vjs-audio-button {
     display: none;
 }
+// only animate when showing because it can be processor heavy
+.vjs-ad-playing.vjs-ad-loading .vjs-loading-spinner:before,
```
incompl: I believe you want `vjs-ad-loading` not `vjs-ad-playing`.

Besides that this just needs a comment explaining where the code is copied from, and what we want to do for a long-term fix.

incompl:
LGTM

bcsdasneogi:
LGTM

PR #310, opened 2017-12-21
Snapshot comments!

incompl:


src/snapshot.js
```diff
@@ -118,15 +121,20 @@ export function restorePlayerSnapshot(player, snapshotObject) {
           currentTime = player.currentTime();
         }
         player.currentTime(currentTime);
+        player.play();
```
ldayananda: this seems like a significant code change 🤔

incompl: This PR was from awhile ago and I don't remember why this is here so I'm removing it. Just don't want to lose the comments.

PR #315, opened 2018-01-02
Ad state refactor

incompl:


src/plugin.js
```diff
@@ -301,435 +264,65 @@ const contribAdsPlugin = function(options) {
...
-
-          this._contentEnding = false;
-          this._contentHasEnded = true;
-          this.state = 'postroll?';
+    debug(...args) {
+      if (this.settings.debug) {
+        if (args.length === 1 && typeof args[0] === 'string') {
+          videojs.log('ADS: ' + args[0]);
+        } else {
+          videojs.log('ADS:', ...args);
```
ldayananda: also is it actually necessary to split this up? won't this work for one argument?

src/plugin.js
```diff
@@ -772,56 +365,21 @@ const contribAdsPlugin = function(options) {
...
-
-    // Events emitted by integrations
-    'adsready',
-    'adserror',
-    'adscanceled',
-    'nopreroll'
-
-  ]), processEvent);
+  // Event handling for the current state.
+  player.on([
```
ldayananda: I believe `VIDEO_EVENTS` is still defined in this file, is it unused now? Also will this change cause html5 events that are currently getting redispatched now to not be handled? If so, is that a change you were intending to make?

src/redispatch.js
```diff
@@ -29,9 +29,7 @@ const cancelEvent = (player, event) => {
 const prefixEvent = (player, prefix, event) => {
   cancelEvent(player, event);
   player.trigger({
-    type: prefix + event.type,
-    state: player.ads.state,
-    originalEvent: event
+    type: prefix + event.type
```
ldayananda: we should probably continue to send the originalEvent

src/redispatch.js
```diff
@@ -70,7 +68,7 @@ const handlePlaying = (player, event) => {
 const handleEnded = (player, event) => {
   if (player.ads.isInAdMode()) {
 
-    // The true ended event fired by plugin.js either after the postroll
+    // The true ended event fired either after the postroll
     // or because there was no postroll.
     if (player.ads.isContentResuming()) {
```
ldayananda: is this the sign that this is the true `ended` event?

test/karma.conf.js
```diff
@@ -23,7 +25,7 @@ module.exports = function(config) {
 
   config.set({
     basePath: '..',
-    frameworks: ['qunit', 'detectBrowsers'],
+    frameworks: ['qunit'],
```
ldayananda: why was detectBrowsers removed? This is still useful for testing locally

src/states/abstract/State.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+  onNoPreroll() {}
+  onNoPostroll() {}
+
+  /*
+   * Method handlers. Different states can override these to provide behaviors.
+   */
+  startLinearAdMode() {
+    videojs.log.warn('Unexpected startLinearAdMode invocation ' +
+      '(State via ' + this.constructor.name + ')');
```
ldayananda: I would expect a message like `Unexpected startLinearAdMode invocation (via ContentPlayback state)` personally

src/states/abstract/State.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+  skipLinearAdMode() {
+    videojs.log.warn('Unexpected skipLinearAdMode invocation ' +
+      '(State via ' + this.constructor.name + ')');
+  }
+
+  /*
+   * Overridden by ContentState and AdState. Should not be overriden elsewhere.
+   */
+  isAdState() {}
```
ldayananda: should this return a default boolean? Seems incorrect to not have anything returned

src/states/abstract/AdState.js
```diff
@@ -0,0 +1,57 @@
...
+  /*
+   * Overrides State.isAdState
+   */
+  isAdState() {
+    return true;
+  }
+
+  /*
+   * We end the content-resuming process on the playing event because this is the exact
+   * moment that ad playback is no longer blocked by ads.
```
ldayananda: ~~ad playback~~ -> content playback

src/states/BeforePreroll.js
```diff
@@ -0,0 +1,72 @@
...
+   * content playback is blocked by the ad plugin.
+   */
+  onPlay(player) {
+    player.ads.debug('Received play event (BeforePreroll)');
+
+    // Don't start content playback yet
+    cancelContentPlay(player);
+
+    // Check for prerolls
+    this.transitionTo(Preroll, this.adsReady);
```
ldayananda: should we even go into Preroll if `nopreroll` is seen already?

src/states/Preroll.js
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  /*
+   * Don't allow the content to start playing while we're dealing with ads.
+   */
+  onPlay(player) {
+    player.ads.debug('Received play event (Preroll)');
+
+    if (!this.inAdBreak() && !this.isContentResuming()) {
```
ldayananda: should this be `!isInAdMode()`?

src/states/Preroll.js
```diff
@@ -0,0 +1,221 @@
...
+      cancelContentPlay(this.player);
+    }
+  }
+
+  /*
+   * adscanceled cancels all ads for the source. Play content now.
+   */
+  onAdsCanceled(player) {
```
ldayananda: this seems like a value that should be passed along to subsequent ad breaks

src/states/Preroll.js
```diff
@@ -0,0 +1,221 @@
...
+
+  /*
+   * An ad error occured. Play content instead.
+   */
+  onAdsError(player) {
+    videojs.log('adserror (Preroll)');
+    // In the future, we may not want to do this automatically.
+    // Integrations should be able to choose to continue the ad break
+    // if there was an error.
+    if (this.inAdBreak()) {
```
ldayananda: should this also be `isInAdMode`?

src/states/ContentPlayback.js
```diff
@@ -0,0 +1,54 @@
...
+/*
+ * This state represents content playback the first time through before
+ * content ends. After content has ended once, we check for postrolls and
+ * move on to the AdsDone state rather than returning here.
+ */
+export default class ContentPlayback extends ContentState {
+
+  init(player) {
+    // Deprecated event.
+    player.trigger({
```
ldayananda: can we remove this now?

src/states/Midroll.js
```diff
@@ -0,0 +1,52 @@
...
+    if (this.inAdBreak()) {
+      endAdBreak(player);
+      this.contentResuming = true;
+    }
+  }
+
+  /*
+   * End midroll break if there is an error.
+   */
+  onAdsError(player) {
```
ldayananda: I believe we should have an `adskip` handler for midrolls too?

src/states/Postroll.js
```diff
@@ -0,0 +1,139 @@
...
+    if (!player.ads.nopostroll_) {
+      player.addClass('vjs-ad-loading');
+
+      this._postrollTimeout = player.setTimeout(function() {
+        player.trigger('adtimeout');
+      }, player.ads.settings.postrollTimeout);
+
+    // No postroll, ads are done
+    } else {
+      player.setTimeout(() => {
```
ldayananda: Seems to me that this could be a state transition and then just trigger the `ended` event. Thoughts?

src/states/Postroll.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  /*
+   * Start the postroll if it's not too late.
+   */
+  startLinearAdMode() {
+    const player = this.player;
+
+    if (!player.ads.isAdPlaying() && !this.contentResuming) {
```
ldayananda: old name `isAdPlaying` should be changed, and use the method for content resuming too

src/states/Postroll.js
```diff
@@ -0,0 +1,139 @@
...
+    } else {
+      videojs.log.warn('Unexpected startLinearAdMode invocation (Postroll)');
+    }
+  }
+
+  /*
+   *
+   */
+  onAdStarted(player) {
+    player.removeClass('vjs-ad-loading');
```
ldayananda: can this go in the AdState class?

src/states/Postroll.js
```diff
@@ -0,0 +1,139 @@
...
+
+  onNoPostroll(player) {
+    if (!this.contentResuming && !player.ads.inAdBreak()) {
+      this.transitionTo(AdsDone);
+    } else {
+      videojs.log.warn('Unexpected nopostroll event (Postroll)');
+    }
+  }
+
+  abort() {
```
ldayananda: Any reason why this is so different to the other AdState subclasses?

src/states/Postroll.js
```diff
@@ -0,0 +1,139 @@
...
+  onContentUpdate(player) {
+    if (this.contentResuming) {
+      this.transitionTo(BeforePreroll);
+    } else if (!player.ads.inAdBreak()) {
+      this.transitionTo(Preroll);
+    }
+  }
+
+  onNoPostroll(player) {
+    if (!this.contentResuming && !player.ads.inAdBreak()) {
```
ldayananda: another place to switch to the methods

src/states/AdsDone.js
```diff
@@ -0,0 +1,19 @@
...
+    // From now on, `ended` events won't be redispatched
+    player.ads._contentHasEnded = true;
+  }
+
+  /*
+   * Midrolls do not play after ads are done.
+   */
+  startLinearAdMode() {
+    videojs.log.warn('Unexpected startLinearAdMode invocation (AdsDone)');
+  }
```
ldayananda: It logically makes sense to trigger `ended` in this state to me...what do you think?

README.md
```diff
@@ -66,7 +66,7 @@ And here are the interaction points you use to send information to the ads plugi
 * `adplaying` (EVENT) - Trigger this event when an ads starts playing. If your integration triggers `playing` event when an ad begins, it will automatically be redispatched as `adplaying`.
 * `adscanceled` (EVENT) — Trigger this event after starting up the player or setting a new video to skip ads entirely. This event is optional; if you always plan on displaying ads, you don't need to worry about triggering it.
 * `adserror` (EVENT) - Trigger this event to indicate that an error in the ad integration has ocurred and any ad states should abort so that content can resume.
-* `nopreroll` (EVENT) - Trigger this event to indicate that there will be no preroll ad. Otherwise, the player will wait until a timeout occurs before playing content. This event is optional, but can improve user experience.
+* `nopreroll` (EVENT) - Trigger this event to indicate that there will be no preroll ads to skip the preroll process completely. This event is optional, but can improve user experience. It is ignored if it is seen after `readyforpreroll`.
```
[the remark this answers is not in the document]
incompl: Looks like something got pasted wrong. Fixing

README.md
```diff
@@ -437,9 +437,9 @@ player.src('movie-high.mp4');
 ### disableNextSnapshotRestore
 Prevents videojs-contrib-ads from restoring the previous video source
 
-If you need to change the video source during ad playback, you can use _disableNextSnapshotRestore_ to prevent videojs-contrib-ads to restore to the previous video source.
+If you need to change the video source during an ad break, you can use _disableNextSnapshotRestore_ to prevent videojs-contrib-ads to restore to the previous video source.
```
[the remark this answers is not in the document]
incompl: Making this change

README.md
```diff
@@ -437,9 +437,9 @@ player.src('movie-high.mp4');
...
 ```js
-if (player.ads.state  === 'ad-playback') {
+if (player.ads.inAdBreak()) {
```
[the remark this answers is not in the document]
incompl: Updating docs

example/example-integration.js
```diff
@@ -100,7 +100,7 @@
     }
 
     player.on('contentended', function() {
-      if (!state.postrollPlayed && player.ads.state === 'postroll?' && playPostroll) {
+      if (!state.postrollPlayed && playPostroll) {
```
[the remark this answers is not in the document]
incompl: No, contentended cannot happen in ad mode.

scripts/umd.rollup.config.js
```diff
@@ -40,7 +40,7 @@ export default {
         }]
       ],
       plugins: [
-        // 'external-helpers',
+        'external-helpers',
```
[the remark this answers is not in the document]
incompl: Yes, Pat suggested it was probably not commented out on purpose.

src/plugin.js
```diff
@@ -3,34 +3,16 @@ This main plugin file is responsible for integration logic and enabling the feat
...
 */
 
-import window from 'global/window';
-
 import videojs from 'video.js';
 
 import redispatch from './redispatch.js';
-import * as snapshot from './snapshot.js';
 import initializeContentupdate from './contentupdate.js';
-import cancelContentPlay from './cancelContentPlay.js';
```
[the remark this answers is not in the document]
incompl: They are depended on by the files that use them, since a lot of stuff has been moved out of plugin.js to other files.

src/plugin.js
```diff
@@ -115,27 +97,23 @@ const contribAdsPlugin = function(options) {
...
-    if (!player.ads.isAdPlaying()) {
+    if (!player.ads.inAdBreak()) {
       return;
     }
 
     player.play();
   });
 
   player.on('nopreroll', function() {
+    player.ads.debug('Received nopreroll event');
```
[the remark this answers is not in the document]
incompl: Yup!

src/plugin.js
```diff
@@ -155,7 +133,7 @@ const contribAdsPlugin = function(options) {
 
   // Replace the plugin constructor with the ad namespace
   player.ads = {
-    state: 'content-set',
+    settings,
```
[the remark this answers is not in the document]
incompl: Which part? Adding settings to the global namespace?

src/plugin.js
```diff
@@ -203,34 +180,20 @@ const contribAdsPlugin = function(options) {
...
-        // duration on adPlayer for an empty ad but we remove it here just to make sure
-        player.removeClass('vjs-ad-loading');
-      }
+      player.ads._state.endLinearAdMode();
     },
 
     // Call this when an ad response has been received but there are no
     // linear ads to be played (i.e. no ads available, or overlays).
-    // This has no effect if we are already in a linear ad mode.  Always
+    // This has no effect if we are already playing an ad.  Always
```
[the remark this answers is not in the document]
incompl: changing it to "in an ad break"

src/plugin.js
```diff
@@ -203,34 +180,20 @@ const contribAdsPlugin = function(options) {
...
     // use endLinearAdMode() to exit from linear ad-playback state.
```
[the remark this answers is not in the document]
incompl: It's suggesting what to do instead of calling skipLinearAdMode during an ad break

src/plugin.js
```diff
@@ -301,435 +264,65 @@ const contribAdsPlugin = function(options) {
...
-            return;
-          }
-
-          this._contentEnding = false;
-          this._contentHasEnded = true;
-          this.state = 'postroll?';
+    debug(...args) {
+      if (this.settings.debug) {
+        if (args.length === 1 && typeof args[0] === 'string') {
+          videojs.log('ADS: ' + args[0]);
```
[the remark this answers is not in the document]
incompl: We can consider that in the future, I don't want to introduce a videojs version dependency here.

src/plugin.js
incompl: In safari it shows each argument on a different line, and this makes simple "log a string" cases look a lot better in safari.

src/plugin.js
incompl: VIDEO_EVENTS is used for redispatch. This change does not impact redispatch.

src/redispatch.js
incompl: I removed both `triggerevent` and `originalEvent` after careful consideration. My conclusion is that they are legacy debug code that should not remain part of the public interface.

src/redispatch.js
incompl: Yup

src/states/AdsDone.js
incompl: I'm not sure it's safe to do that. For example, one of the ended events happens in a 1 millisecond timeout, but not others. I think this is a good area of future improvement, but I don't think it can be safely changed in this PR without lots of re-testing.

src/states/BeforePreroll.js
incompl: Maybe, but I'm not sure what the implications are. I'd put this on the "something to consider in the future" list rather than re-test.

src/states/ContentPlayback.js
incompl: Yeah let's go for it

src/states/Midroll.js
incompl: I don't think so, adskip has to happen before the ad break starts, and in midrolls the ad break starts right away.

src/states/Postroll.js
incompl: To summarize previous comment: not sure what the implications are, want to push to possible future work rather than retest.

src/states/Postroll.js
incompl: Updating

src/states/Postroll.js
incompl: `vjs-ad-loading` management is very different for prerolls and postrolls, so to keep it as clear as possible I isolated it to those classes rather than trying to share code.

incompl: I am updating the empty comment here though.

src/states/Postroll.js
incompl: Updating

src/states/Postroll.js
incompl: It just evolved to be this way based on what this class was doing. I'm open to other ideas, but this doesn't seem problematic per se.

src/states/Preroll.js
incompl: isAdBreak is the new isInAdMode. Are you saying you don't think the content resuming check should be there? That has historically been what we've done, and I'm not aware of a reason to change it: I don't think you should pause the player during content resuming.

src/states/Preroll.js
incompl: No, inAdbreak is the new one.

src/states/abstract/AdState.js
incompl: Good catch

src/states/abstract/State.js
incompl: I wanted to include the name of the file that the log message is in to make it easier to debug.

src/states/abstract/State.js
incompl: If anything, I think the default should be to throw an error. My intention was to make it clear that it must be overwritten.

test/karma.conf.js
incompl: Because it picks up Canary and potentially browsers that the tests don't work in, which can result in false positives. Maybe there is a better solution, but false positives can result in a lot of lost time.

README.md
```diff
@@ -115,9 +115,11 @@ Returns true if content is resuming after an ad. This is part of ad mode.
...
 #### isAdPlaying()
 
-Returns true if a linear ad is playing. This is part of ad mode.
-This relies on `startLinearAdMode` and `endLinearAdMode` because that is the
-most authoritative way of determinining if an ad is playing.
+Deprecated. Does the same thing as `inAdBreak` but has a misleading name. videojs-contrib-ads does not know when an ad is actually playing, that's up to the integration. However, ads can only play during an ad break.
+
+#### inAdBreak()
+
+This method returns true during the time between startLinearAdMode and endLinearAdMode where an integratino may play ads. This is part of ad mode.
```
ldayananda: Typo integratino

src/plugin.js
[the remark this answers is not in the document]
ldayananda: Yeah, and that state is no longer available and should not have been relied on

src/plugin.js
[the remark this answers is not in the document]
ldayananda: Ah, sorry missed the 2 lines above this originally

src/plugin.js
ldayananda: might make sense to use `String.concat` instead then 🤔

src/redispatch.js
ldayananda: Hmm, let's leave it like this(removed) for now and discuss again before releasing

src/states/Postroll.js
ldayananda: Not necessarily problematic, it just seemed to me at first glance that most of what is in `abort ()` can go in `cleanup()`. Except for the contentResuming change and triggering ended, I suppose.

src/states/Preroll.js
ldayananda: no, I think the `!isContentResuming()` is likely to be correct. I think the `!inAdBreak()` part is a bit confusing since plays will not be cancelled while waiting for `startLinearAdMode()` while in the Preroll state, right?

ldayananda: uh, sorry I meant plays will not be cancelled while the ad is actually playing

src/states/Preroll.js
ldayananda: I'm on board with removing it

src/states/abstract/State.js
ldayananda: Maybe I wasn't clear, I just meant having state be last in the message is clearer 😆 
`videojs.log.warn('Unexpected startLinearAdMode invocation (via', this.constructor.name, ' state)')`

src/states/abstract/State.js
ldayananda: I like the idea of throwing an error

test/karma.conf.js
ldayananda: Though I understand your frustration, I don't believe reducing tested browsers helps us in the long run. Having the tests run on more browsers is a good way to ensure browsers we don't normally develop in are tested. 

An alternative to what you did here would be to remove browsers that we know the tests don't run in from the list of browsers that were detected. I'll be happy to dig up an example of how to do that if needed

incompl: It's not an emotional choice, it's pragmatism: if everyone is running the tests differently, our results are not scientific. If we want to run the tests on more platforms, I would suggest we add them explicitly rather than saying "whatever happens to be installed"

incompl: Additionally, any browsers we add should be confirmed to work with the tests before committing.

ldayananda: the tests will then fall for contributors that don't have those explicit browsers installed

example/example-integration.js
```diff
@@ -92,7 +92,7 @@
     // initialize the ads plugin, passing in any relevant options
     player.ads(options);
 
-    // request ad inventory whenever the player gets new content to play
+    // request ad inventory whenever the player gets content to play
```
ldayananda: maybe this listener should be updated to listen to contentchanged?

test/karma.conf.js
incompl: That seems like a good behavior if your system can't run the full test suite.

src/states/Midroll.js
```diff
@@ -17,7 +17,6 @@ export default class Midroll extends AdState {
     const player = this.player;
 
     if (!this.inAdBreak() && !this.isContentResuming()) {
-      player.ads.debug('startLinearAdMode (Midroll)');
```
ldayananda: why remove this log?

README.md
incompl: Fixing

src/states/Postroll.js
```diff
@@ -43,7 +43,6 @@ export default class Postroll extends AdState {
     const player = this.player;
 
     if (!player.ads.isAdPlaying() && !this.contentResuming) {
-      player.ads.debug('startLinearAdMode (Postroll)');
```
ldayananda: why remove this log?

src/states/Postroll.js
incompl: I think ideally it would all go in cleanup, but I'm hesitant to change the `ended` event more than necessary, just to keep the risk of this PR to a minimum. Definitely something to consider in the future.

src/states/abstract/State.js
```diff
@@ -106,8 +106,8 @@ export default class State {
       this.onAdTimeout(player);
     } else if (type === 'ads-ad-started') {
       this.onAdStarted(player);
-    } else if (type === 'contentupdate') {
-      this.onContentUpdate(player);
+    } else if (type === 'contentchanged') {
```
ldayananda: just for confirmation: contentupdate will no longer be used for state transitions?

src/states/Preroll.js
incompl: The historical behavior has been that plays are not cancelled during an ad break. I believe this works out because the player controls are hidden, so the user cannot request a content play. A programatic call to play during ads would break things. Maybe we can improve this using middleware!

src/states/Preroll.js
incompl: Cool. Again, to limit the scope of this PR, I'd like to do that in a separate PR.

src/states/Midroll.js
incompl: I had added it in a relatively recent commit and found it's basically redundant with the "Starting ad break" log.

src/states/Postroll.js
incompl: I had added it in a relatively recent commit and found it's basically redundant with the "Starting ad break" log.

src/states/abstract/State.js
incompl: Correct

example/example-integration.js
incompl: Originally I didn't change it because it seems to run on the initial contentupdate. Now that I look at it again it seems like it might request ads twice sometimes as it is? I am going to update it.

test/test.ads.js
```diff
@@ -325,299 +268,160 @@ QUnit.test('changing src does not trigger "contentupdate" during ad playback', f
...
+  assert.strictEqual(setTimeoutSpy.callCount, 2, 'two timers were created (`cancelPlayTimeout` and `_prerollTimeout`)');
+  assert.ok(timerExists(this, this.player.ads.cancelPlayTimeout), '`cancelPlayTimeout` exists');
+  assert.ok(timerExists(this, this.player.ads._state._timeout), 'preroll timeout exists');
 
-  this.player.trigger('adserror');
-  assert.strictEqual(this.player.ads.state, 'content-playback');
+  this.clock.tick(1);
+  this.player.ads.startLinearAdMode();
+  this.player.ads.endLinearAdMode();
+  assert.notOk(timerExists(this, this.player.ads.cancelPlayTimeout), '`cancelPlayTimeout` was canceled');
```
ldayananda: test message seems wrong - `play was canceled` ? Or do we want the timeout to not exist?

test/test.ads.js
```diff
@@ -325,299 +268,160 @@ QUnit.test('changing src does not trigger "contentupdate" during ad playback', f
...
   this.player.trigger('ended');
-  assert.strictEqual(this.player.ads.state, 'postroll?');
-
   this.player.trigger('adserror');
-  assert.strictEqual(this.player.ads.state, 'content-resuming');
-  assert.strictEqual(this.player.ads.triggerevent, 'adserror', 'adserror should be the trigger event');
 
-  this.clock.tick(1);
-  assert.strictEqual(this.player.ads.state, 'content-playback');
+  this.clock.tick(1000);
```
ldayananda: this can go back to `1` now, right?

test/test.ads.js
```diff
@@ -325,299 +268,160 @@ QUnit.test('changing src does not trigger "contentupdate" during ad playback', f
...
   this.player.trigger('ended');
-  assert.strictEqual(this.player.ads.state, 'postroll?');
-
   this.player.trigger('adtimeout');
-  assert.strictEqual(this.player.ads.state, 'content-resuming');
-  assert.strictEqual(this.player.ads.triggerevent, 'adtimeout', 'adtimeout should be the trigger event');
 
-  this.clock.tick(1);
-  assert.strictEqual(this.player.ads.state, 'content-playback');
+  this.clock.tick(1000);
```
ldayananda: same here

test/test.ads.js
```diff
@@ -325,299 +268,160 @@ QUnit.test('changing src does not trigger "contentupdate" during ad playback', f
...
-  assert.strictEqual(this.player.ads.state, 'content-resuming');
-  assert.strictEqual(this.player.ads.triggerevent, 'adskip', 'adskip should be the trigger event');
-
-  this.clock.tick(1);
-  assert.strictEqual(this.player.ads.state, 'content-playback');
+  this.player.trigger('adtimeout'); // preroll times out
+  this.player.trigger('ended'); // content ends (contentended)
+  this.player.trigger('adskip'); // postroll skipped
+  
+  this.clock.tick(1000);
```
ldayananda: also here

test/test.ads.js
```diff
@@ -325,299 +268,160 @@ QUnit.test('changing src does not trigger "contentupdate" during ad playback', f
...
 
-  this.clock.tick(1000);
-  assert.strictEqual(endedSpy.callCount, 1, 'we should have fired ended from the timeout');
+  this.player.ads.startLinearAdMode(); // start postroll
+  this.player.ads.endLinearAdMode();
+  assert.strictEqual(endedSpy.callCount, 1, 'ended event happened');
 });
 
-QUnit.test('an "ended" event is not fired in "content-resuming" via a timeout if fired naturally', function(assert) {
+QUnit.test('ended events', function(assert) {
```
ldayananda: we should expand this test name

test/test.ads.js
```diff
@@ -325,299 +268,160 @@ QUnit.test('changing src does not trigger "contentupdate" during ad playback', f
...
-  spy = sinon.spy();
-  this.player.on('adend', spy);
-  this.player.trigger('adserror');
-  assert.strictEqual(this.player.ads.state, 'content-resuming');
-  assert.strictEqual(this.player.ads.triggerevent, 'adserror', 'The reason for content-resuming should have been adserror');
 
-  this.player.trigger('playing');
-  assert.strictEqual(this.player.ads.state, 'content-playback');
-  assert.strictEqual(spy.getCall(0).args[0].type, 'adend', 'adend should be fired when we enter content-playback from adserror');
+  assert.strictEqual(adendSpy.callCount, 0, 'no adend yet');
```
ldayananda: what is this test testing now?

test/test.ads.js
```diff
@@ -748,16 +524,34 @@ QUnit.test('player events during prerolls are prefixed if tech is reused for ad'
...
+  assert.strictEqual(sawPlaying.callCount, 0, 'no playing fired');
+  assert.strictEqual(sawPause.callCount, 0, 'no pause fired');
+  assert.strictEqual(sawEnded.callCount, 0, 'no ended fired');
+  assert.strictEqual(sawFirstplay.callCount, 0, 'no firstplay fired');
+  assert.strictEqual(sawLoadedalldata.callCount, 0, 'no loadedalldata fired');
+  assert.strictEqual(sawAdloadstart.callCount, 1, 'adloadstart fired');
+  assert.strictEqual(sawAdpause.callCount, 1, 'adpause fired');
+  assert.strictEqual(sawAdended.callCount, 1, 'adended fired');
+  assert.strictEqual(sawAdfirstplay.callCount, 1, 'adfirstplay fired');
+  assert.strictEqual(sawAdloadedalldata.callCount, 1, 'adloadedalldata fired');
```
ldayananda: I think that `firstplay`, `loadstart`, `playing`, `loadedalldata`, `pause`, `ended` were meant to be checked as well to make sure that they were canceled?

test/test.ads.js
```diff
@@ -889,85 +684,115 @@ QUnit.test('startLinearAdMode should only trigger adstart from correct states',
...
+  this.player.ads.endLinearAdMode();
+  assert.strictEqual(adstart.callCount, 3, 'Ads done');
+  
 });
 
 QUnit.test('ad impl can notify contrib-ads there is no preroll', function(assert) {
 
-  this.player.ads.state = 'preroll?';
+  this.player.trigger('loadstart');
+  this.player.trigger('nopreroll');
```
ldayananda: what happens if nopreroll is triggered after the first of `play` or `adsready`? i.e loadstart -> play -> nopreroll

test/test.ads.js
```diff
@@ -889,85 +684,115 @@ QUnit.test('startLinearAdMode should only trigger adstart from correct states',
...
-    }
-  };
   this.player.on('ended', ended);
-  this.player.ads.state = 'content-playback';
+
   this.player.trigger('contentended');
 
   this.clock.tick(10000);
-
   assert.ok(ended.calledOnce, 'Ended triggered');
```
ldayananda: seems like this is a duplicate of the previous test now

test/test.ads.js
```diff
@@ -1062,7 +887,8 @@ QUnit.test('Check incorrect addition of vjs-live during ad-playback', function(a
 
 QUnit.test('Check for existence of vjs-live after ad-end for LIVE videos',
   function(assert) {
-    this.player.trigger('adstart');
+    this.player.trigger('adsready');
```
ldayananda: loadstart?

src/redispatch.js
incompl: After discussion adding back just originalEvent

src/states/abstract/State.js
incompl: Oh, I wrote "State" in there as the name of the file, because that is what this.constructor.name would be.

src/states/abstract/State.js
incompl: Doing that

test/karma.conf.js
incompl: After discussion we're changing it to detectBrowsers with a whitelist.

src/states/abstract/State.js
ldayananda: yeah the order of the words. What do you mean by "ClassName via ClassName"?

incompl: In these log messages I've been putting in parens `ClassNameThatLoggedThisMessage` and for this exception the intended reading is `ClassNameThatLoggedThisMessage by way of ConcreteClassThatInheritedTheBehavior`

test/test.ads.js
```diff
@@ -1,15 +1,9 @@
-var timerExists = function(env, keyOrId) {
-  var timerId = _.isNumber(keyOrId) ? keyOrId : env.player.ads[String(keyOrId)];
-  return env.clock.timers.hasOwnProperty(String(timerId));
+var timerExists = function(env, id) {
+  return env.clock.timers.hasOwnProperty(id);
 };
 
 QUnit.module('Ad Framework', window.sharedModuleHooks());
 
-QUnit.test('begins in content-set', function(assert) {
```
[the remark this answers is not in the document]
incompl: can't hurt

test/test.ads.js
```diff
@@ -47,29 +41,26 @@ QUnit.test('pauses to wait for prerolls when the plugin loads AFTER play', funct
 QUnit.test('stops canceling play events when an ad is playing', function(assert) {
   var setTimeoutSpy = sinon.spy(window, 'setTimeout');
```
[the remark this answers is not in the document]
incompl: player.setTimeout ultimately calls window.setTimeout so it ends up being the same. I think a better improvement for this test would be to test the functionality more directly. The test description says "stops canceling play events when an ad is playing". A better test would start an ad playing and see if play events get canceled. There are tons of improvements that could be made to these tests, so I had to pick my battles as I updated these.

test/test.ads.js
```diff
@@ -47,29 +41,26 @@ QUnit.test('pauses to wait for prerolls when the plugin loads AFTER play', funct
...
   this.player.trigger('play');
-  assert.strictEqual(setTimeoutSpy.callCount, 2, 'two timers were created (`cancelPlayTimeout` and `adTimeoutTimeout`)');
-  assert.ok(timerExists(this, 'cancelPlayTimeout'), '`cancelPlayTimeout` exists');
-  assert.ok(timerExists(this, 'adTimeoutTimeout'), '`adTimeoutTimeout` exists');
+  assert.strictEqual(setTimeoutSpy.callCount, 2, 'two timers were created (`cancelPlayTimeout` and `_prerollTimeout`)');
+  assert.ok(timerExists(this, this.player.ads.cancelPlayTimeout), '`cancelPlayTimeout` exists');
+  assert.ok(timerExists(this, this.player.ads._state._timeout), 'preroll timeout exists after play');
 
   this.player.trigger('adsready');
-  assert.strictEqual(setTimeoutSpy.callCount, 3, '`adTimeoutTimeout` was re-scheduled');
```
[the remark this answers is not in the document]
incompl: Yes, the implementation changed without the behavior changing and this assertion started failing without an underlying defect in the code. I removed it to help our tests focus more on tests of functionality and reduce false positives.

test/test.ads.js
incompl: It's saying that the timeout named "cancelPlayTimeout" was canceled, confirmed by the fact that the variable saving the timeout ID was deleted.

src/states/abstract/State.js
ldayananda: 💡  Ah I see. Makes sense

test/test.ads.js
incompl: Yeah changing it back

test/test.ads.js
incompl: updated

test/test.ads.js
incompl: Yes, as long as I also update it to call `this.player.ads.skipLinearAdMode();` rather than trigger adskip.

incompl: I guess it was timing out rather than skipping as it looked like.

test/test.ads.js
incompl: Updated

test/test.ads.js
incompl: Updated this test to do what it says it should

test/test.ads.js
incompl: Looks like those are there, above this line.

test/test.ads.js
incompl: Added two more tests to cover these variants.

test/test.ads.js
ldayananda: ah interesting

test/test.ads.js
ldayananda: Ah missed that 🆒

test/test.ads.js
incompl: removing the duplicate test

test/test.ads.js
incompl: Updating

src/plugin.js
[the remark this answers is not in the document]
incompl: Yeah the state part is mentioned. The settings part isn't in the migration guide since it's a new thing and doesn't require migration.

src/plugin.js
incompl: They might not all be strings, though.

ldayananda: `Array.join` perhaps?

incompl: That would still do string concatenation and bypass the browser's special console output for non-string types.

ldayananda: isn't that the same for `'ADS: ' + args[0]` ?

incompl: It does a type check on line 304 to cover that.

test/test.events-no-postroll.js
```diff
@@ -32,9 +32,10 @@ QUnit.module('Final Events With No Postroll', {
 
 QUnit.test('final ended event with no postroll: just 1', function(assert) {
   var done = assert.async();
-
   var endedEvents = 0;
 
+  this.player.ads.settings.postrollTimeout = 1;
```
ldayananda: why was this necessary?

README.md
```diff
@@ -38,6 +54,12 @@ videojs('video', {}, function() {
 You may also use the Javascript and CSS links from the following to get started:
 [https://cdnjs.com/libraries/videojs-contrib-ads](https://cdnjs.com/libraries/videojs-contrib-ads)
 
+### Using a module system
+
+If you are loading `videojs-contrib-ads` using modules, do this:
+
+TODO
```
ldayananda: This TODO still meant to be here?

README.md
```diff
@@ -51,43 +73,51 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
+it is initialized. This can happen before or after the `play` event.
+* Contrib Ads triggers `readyforpreroll` (EVENT) -- This event is fired after both `play` and `adsready` have ocurred.
+This signals that the integration may begin an ad break by calling `startLinearAdMode`.
+* Integration calls `player.ads.startLinearAdMode()` (METHOD) -- This begins an ad break. During this time, your integration
+plays ads. videojs-contrib-ads does not handle actual ad playback.
+* Integration triggers `ads-ad-started` (EVENT) - Trigger this when each individual ad begins. This removes the loading spinner, which otherwise stays up during the ad break. It's possible for an ad break
+to end without an ad starting, in which case the spinner stays up the whole time.
+* Integration calls `player.ads.endLinearAdMode()` (METHOD) -- This ends an ad break. As a result, content will play.
+* Content plays.
+* To play a Midroll ad, start an end an ad break with `player.ads.startLinearAdMode()` and `player.ads.endLinearAdMode()` at any time during content playback.
```
ldayananda: Typo ~~start an end an ad break~~ start and end an ad break

README.md
```diff
@@ -51,43 +73,51 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
+* Contrib Ads triggers `contentended` (EVENT) -- This event means that it's time to play a postroll ad.
+* To play a Postroll ad, start an end an ad break with `player.ads.startLinearAdMode()` and `player.ads.endLinearAdMode()`.
```
ldayananda: same here

README.md
```diff
@@ -51,43 +73,51 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
+* Contrib Ads triggers `ended` (EVENT) -- This standard media event happens when all ads and content have completed. After this, no additional ads are expected, even if the user seeks backwards.
+
+This is the basic flow for a simple use case, but there are other things the integration can do:
+
+* `skipLinearAdMode` (METHOD) -- At a time when `startLinearAdMode` is expected, calling `skipLinearAdMode` will immediately resume content playback instead.
+* `nopreroll` (EVENT) -- You can trigger this event even before `readyforpreroll` to indicate that no preroll will play. The ad plugin will not check for prerolls and will instead immediately begin content playback after the `play` event (or immidiately, if playback was already requested).
```
ldayananda: Typo `immidiately`.

Also I would delete the first immediately in this sentence

README.md
```diff
@@ -51,43 +73,51 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
+* `nopostroll` (EVENT) -- Similar to `nopreroll`, you can trigger this event even before `contentended` to indicate that no postroll will play.  The ad plugin will not wait for a postroll to play and will instead immediately trigger the `ended` event.
+* `adserror` (EVENT) -- This event skips prerolls when seen before a preroll ad break. It skips postrolls if called after contentended and before a postroll ad break. It ends linear ad mode if seen during an ad break.
```
ldayananda: We should say when the integration should trigger this event

README.md
```diff
@@ -51,43 +73,51 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
- * `readyforpreroll` (EVENT) — Fires when a content video is about to play for the first time, so your integration can indicate that it wants to play a preroll.
+There are some other useful events that videojs-contrib-ads may trigger:
 
-Note: A `contentplayback` event is sent but should not be used as it is being removed. The `playing` event has the same meaning and is far more reliable.
- 
-And here are the interaction points you use to send information to the ads plugin:
+ * `contentchanged` (EVENT) -- Fires when a new content video has been loaded in the player (specially, at the same time as the `loadstart` media event for the new source). This means the ad workflow has restarted from the beginning. Your integration will need to trigger `adsready` again, for example. Note that when changing sources, the playback state of the player is retained: if the previous source was playing, the new source will also be playing and the ad workflow will not wait for a new `play` event.
```
ldayananda: specially? Or specifically?

README.md
```diff
@@ -51,43 +73,51 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
-* `ads.startLinearAdMode()` (METHOD) — Call this method to signal that your integration is about to play a linear ad. This method triggers `adstart` to be emitted by the player.
-* `ads.endLinearAdMode()` (METHOD) — Call this method to signal that your integration is finished playing linear ads, ready for content video to resume. This method triggers `adend` to be emitted by the player.
-* `ads.skipLinearAdMode()` (METHOD) — Call this method to signal that your integration has received an ad response but is not going to play a linear ad.  This method triggers `adskip` to be emitted by the player.
-* `ads.stitchedAds()` (METHOD) — Get or set the `stitchedAds` setting.
-* `ads.videoElementRecycled()` (METHOD) - Returns true if ad playback is taking place in the content element.
+Deprecated events:
 
-In addition, video.js provides a number of events and APIs that might be useful to you.
-For example, the `ended` event signals that the content video has played to completion.
+* `contentplayback` (EVENT) -- The `playing` media event has the same meaning and is far more reliable.
```
ldayananda: Although I like these descriptions, they seem to fit better in the migration guide. I would then link to the most recent migration guide in this section

README.md
```diff
@@ -51,43 +73,51 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
 ### Public Methods
 
-These are methods that can be called at runtime to inspect the ad plugin's state. You do
-not need to implement them yourself.
+These are methods on `player.ads` that can be called at runtime to inspect the ad plugin's state. You do not need to implement them yourself.
 
 #### isInAdMode()
 
-Returns true if player is in ad mode.
+Returns true if the player is in ad mode.
```
ldayananda: didn't this change to be isInAdBreak?

README.md
```diff
@@ -405,13 +397,13 @@ The plugin triggers a number of custom events on the player during its operation
...
 
 ### adend
-The player has returned from linear ad playback mode. This event is fired directly as a consequence of calling `endLinearAdMode()`. Note that multiple ads may have played back between `adstart` and `adend`.
+The player has returned from linear ad playback mode. This event is fired directly as a consequence of calling `endLinearAdMode()`. Note that multiple ads may have played back in the ad break between `adstart` and `adend`.
 
 ### adskip
-The player is skipping a linear ad opportunity and content-playback should resume immediately.  This event is fired directly as a consequence of calling `skipLinearAdMode()`. It can indicate that an ad response was made but returned no linear ad content or that no ad call is going to be made at either the preroll or postroll timeout opportunities.
+The player is skipping a linear ad opportunity and content-playback should resume immediately.  This event is fired directly as a consequence of calling `skipLinearAdMode()`. As examples, it can indicate that an ad response was made but returned no linear ad content or that no ad call is going to be made due to an error.
```
ldayananda: ~~As~~ For example? Also I think you meant ad request?

test/test.events-no-postroll.js
incompl: The test times out otherwise. This is just so the test runs faster basically. I'll add a comment.

README.md
incompl: Yeah, I want to link to the example from https://github.com/videojs/videojs-contrib-ads/pull/312 once that is merged. Was planning to do that in rebase. I'll add more info.

README.md
incompl: 👍

README.md
incompl: 👍

README.md
incompl: 👍

README.md
incompl: I don't think there is a clear answer to that, it depends on if you want the extra behavior or not. I think we need to change how adserror works. For now, just want to make it clear what it does.

README.md
incompl: Updated to specifically

README.md
incompl: Ad mode includes loading ads and content resuming. The ad break is just a part of that.

README.md
incompl: Oh yeah, since we're removing it in this version we can move it from "deprecated" to the migration guide.

README.md
incompl: Updating this

README.md
ldayananda: Ah sorry, keep getting confused

README.md
ldayananda: To clarify, for prerolls and postrolls the break is skipped if it is triggered before ad mode or before the ad break is started? I think that's not clear for prerolls

README.md
```diff
@@ -88,26 +90,25 @@ plays ads. videojs-contrib-ads does not handle actual ad playback.
...
+ * `contentchanged` (EVENT) -- Fires when a new content video has been loaded in the player (specifically, at the same time as the `loadstart` media event for the new source). This means the ad workflow has restarted from the beginning. Your integration will need to trigger `adsready` again, for example. Note that when changing sources, the playback state of the player is retained: if the previous source was playing, the new source will also be playing and the ad workflow will not wait for a new `play` event.
 
 Deprecated events:
 
-* `contentplayback` (EVENT) -- The `playing` media event has the same meaning and is far more reliable.
 * `contentupdate` (EVENT) -- Replaced by `contentchanged`, which is more reliable.
```
ldayananda: I meant for these two lines to move to the migration guide as well

README.md
incompl: It says before the ad break, which is exactly accurate.

README.md
incompl: I'll add a "deprecation" section to the migration guide and will leave this here so there is still a deprecation warning in the readme.

test/states/test.ContentPlayback.js
```diff
@@ -0,0 +1,94 @@
...
+    };
+  }
+});
+
+QUnit.test('only plays on init on correct conditions', function(assert) {
+  this.player.paused = () => false;
+  this.player.ads._cancelledPlay = false;
+  this.player.ads._pausedOnContentupdate = false;
+  this.contentPlayback.init(this.player);
+  assert.equal(this.played, false);
```
ldayananda: this doesn't seem right since `paused() -> false`

ldayananda: maybe `played` should be `playTriggered`

test/states/test.Midroll.js
```diff
@@ -0,0 +1,50 @@
...
+
+QUnit.test('starts an ad break on init', function(assert) {
+  this.midroll.init(this.player);
+  assert.equal(this.player.ads.adType, 'midroll', 'ad type is midroll');
+  assert.equal(this.adBreakStartStub.callCount, 1, 'ad break started');
+});
+
+QUnit.test('ends an ad break on endLinearAdMode', function(assert) {
+  this.midroll.init(this.player);
+  this.player.ads._inLinearAdMode = true;
```
ldayananda: do you have to hardcode this? could we do this as part of the stub?

test/states/test.Postroll.js
```diff
@@ -0,0 +1,152 @@
...
+  assert.equal(this.events[0], 'ended', 'saw ended event');
+});
+
+QUnit.test('no endLinearAdMode on adserror if not in ad break', function(assert) {
+  this.player.ads.endLinearAdMode = sinon.spy();
+
+  this.postroll.init(this.player);
+  this.player.ads.inAdBreak = () => false;
+  this.postroll.onAdsError(this.player);
+  assert.equal(this.player.ads.endLinearAdMode.callCount, 0, 'linear ad mode ended');
```
ldayananda: does ended get called?

test/states/test.Preroll.js
```diff
@@ -0,0 +1,143 @@
...
+});
+
+QUnit.test('plays a preroll (adsready true)', function(assert) {
+  this.preroll.init(this.player, true);
+  assert.equal(this.preroll.adsReady, true, 'adsready from init');
+  assert.equal(this.events[0], 'readyforpreroll', 'readyforpreroll from init');
+  assert.equal(this.preroll.inAdBreak(), false, 'not in ad break');
+
+  this.preroll.startLinearAdMode();
+  this.player.ads._inLinearAdMode = true; // Because adBreak.start is mocked.
```
ldayananda: comments should go on the line above

test/states/test.Preroll.js
```diff
@@ -0,0 +1,143 @@
...
+  assert.equal(this.adBreakStartStub.callCount, 1, 'ad break started');
+  assert.equal(this.player.ads.adType, 'preroll', 'adType is preroll');
```
ldayananda: did we have one of these for postroll?

ldayananda:
Also, this comment got hidden: https://github.com/videojs/videojs-contrib-ads/pull/315#discussion_r168346332

test/states/abstract/test.AdState.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+QUnit.module('AdState', {
+  beforeEach: function() {
+    this.player = {
+      ads: {}
+    };
+
+    this.adState = new AdState(this.player);
+    this.adState.transitionTo = (newState) => {
+      this.transitionTo = newState.name;
```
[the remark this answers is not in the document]
incompl: I was also considering transitionedTo and newState, newState is probably the most obvious

test/states/test.Midroll.js
incompl: I wanted the stub to reflect an "initial" state and have each test advance it forward to the necessary state, rather than having the stub be kind of a random assortment of properties.

incompl:
That link isn't taking me to the right place, which comment is that?

test/states/test.Preroll.js
incompl: adding one

test/states/test.Preroll.js
incompl: changing

test/states/test.Postroll.js
incompl: might as well assert not

incompl: sorry the other way, we assert that ended is still called

test/states/test.ContentPlayback.js
incompl: changing it

test/states/abstract/test.ContentState.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+QUnit.test('handles content changed when playing', function(assert) {
+  this.player.paused = () => false;
+  this.player.pause = sinon.stub();
+
+  this.contentState.onContentChanged(this.player);
+  assert.equal(this.transitionTo, 'Preroll');
+  assert.equal(this.player.pause.callCount, 1, 'paused player');
+  assert.equal(this.player.ads._pausedOnContentupdate, true, 'set _pausedOnContentupdate');
```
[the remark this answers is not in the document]
incompl: it's extra, might as well add it

test/states/test.Midroll.js
ldayananda: I see. Would calling `startLinearAdMode` be equivalent then?

README.md
```diff
@@ -51,42 +75,50 @@ The plugin will emit an error if it detects that it it missed a `loadstart` even
...
+
+This is the basic flow for a simple use case, but there are other things the integration can do:
+
+* `skipLinearAdMode` (METHOD) -- At a time when `startLinearAdMode` is expected, calling `skipLinearAdMode` will immediately resume content playback instead.
+* `nopreroll` (EVENT) -- You can trigger this event even before `readyforpreroll` to indicate that no preroll will play. The ad plugin will not check for prerolls and will instead begin content playback after the `play` event (or immediately, if playback was already requested).
+* `nopostroll` (EVENT) -- Similar to `nopreroll`, you can trigger this event even before `contentended` to indicate that no postroll will play.  The ad plugin will not wait for a postroll to play and will instead immediately trigger the `ended` event.
+* `adserror` (EVENT) -- This event skips prerolls when seen before a preroll ad break. It skips postrolls if called after contentended and before a postroll ad break. It ends linear ad mode if seen during an ad break.
```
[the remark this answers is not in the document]
incompl: Yup

test/states/test.Midroll.js
incompl: Yeah, originally I wanted to limit the test to only calling one function, but looking back on it now, using startLinearAdMode is a more solid test.

incompl: My mistake. startLinearAdMode is not defined in Midroll.js, so that's not an option.

ldayananda: Ah, this.player.ads._inLinearAdMode is always mocked in this test file. I suppose that is fine then

incompl: Ah! Just realized that when you enter the Midroll state, _inLinearAdMode would always be true, so it make sense for this to change in the mock. We've gone full circle on this one.

incompl: Updated

PR #318, opened 2018-01-19
Fix: content can play too early after contentupdate

alex-barstow:
## Description

This addresses a couple of issues surrounding contentupdates:

1.  Content can start playing immediately after a source change because we don't proactively pause the player if it is playing following a `contentupdate`
2. There is no `contentupdate` event handler in the 'preroll?' state, so if one occurs we remain in ‘preroll?’ through the source change. And because the ad [timeout](https://github.com/videojs/videojs-contrib-ads/blob/master/src/plugin.js#L397-L399) isn’t cleared, the eventual `adtimeout` event causes a transition directly into ‘content-playback’, bypassing 'content-set' / 'ads-ready?' / 'ads-ready'

**EDIT: We are going to wait to address no.2 in the upcoming refactor**

## Specific Additions

1. If we are in 'content-playback' and a `contentupdate` occurs, pause the player if it is not already paused. 
2. Add a `contentupdate()` handler in 'preroll?' that changes state to 'content-set'

**As mentioned above: we'll hold off on no.2 until the refactor**

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -139,6 +139,7 @@ const contribAdsPlugin = function(options) {
   // Restart the cancelContentPlay process.
   player.on('playing', () => {
     player.ads._cancelledPlay = false;
+    player.ads._pausedOnContentupdate = false;
   });
 
   player.one('loadstart', () => {
@@ -435,6 +436,9 @@ const contribAdsPlugin = function(options) {
         },
         nopreroll() {
           this.state = 'content-playback';
+        },
+        contentupdate() {
+          this.state = 'content-set';
         }
       }
     },
@@ -638,10 +642,10 @@ const contribAdsPlugin = function(options) {
           triggerevent: player.ads.triggerevent
         });
 
-        // Play the content if cancelContentPlay happened and we haven't played yet.
-        // This happens if there was no preroll or if it errored, timed out, etc.
-        // Otherwise snapshot restore would play.
-        if (player.ads._cancelledPlay) {
+        // Play the content if cancelContentPlay happened or we paused on 'contentupdate'
+        // and we haven't played yet. This happens if there was no preroll or if it
+        // errored, timed out, etc. Otherwise snapshot restore would play.
+        if (player.ads._cancelledPlay || player.ads._pausedOnContentupdate) {
           if (player.paused()) {
             player.play();
           }
@@ -666,6 +670,10 @@ const contribAdsPlugin = function(options) {
           if (player.paused()) {
             this.state = 'content-set';
           } else {
+            // If we get here we missed the 'play' event, so pause the player so
+            // that content doesn't start playing immediately after the source change
+            player.pause();
+            player.ads._pausedOnContentupdate = true;
             this.state = 'ads-ready?';
           }
         },
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -436,9 +436,6 @@ const contribAdsPlugin = function(options) {
         },
         nopreroll() {
           this.state = 'content-playback';
-        },
-        contentupdate() {
-          this.state = 'content-set';
         }
       }
     },
--- a/test/test.ads.js
+++ b/test/test.ads.js
@@ -729,6 +729,34 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
   assert.strictEqual(spy.getCall(0).args[0].type, 'readyforpreroll', 'readyforpreroll should have been triggered.');
 });
 
+QUnit.test('contentupdate in content-playback transitions to content-set if the player is paused', function(assert) {
+  assert.expect(2);
+
+  this.player.trigger('play');
+  this.player.trigger('adtimeout');
+  assert.strictEqual(this.player.ads.state, 'content-playback');
+  this.player.paused = function() {
+    return true;
+  };
+  this.player.trigger('contentupdate');
+  assert.strictEqual(this.player.ads.state, 'content-set');
+});
+
+QUnit.test('contentupdate in content-playback pauses player and transitions to ads-ready? if the player is not paused', function(assert) {
+  assert.expect(3);
+
+  this.player.trigger('play');
+  this.player.trigger('adtimeout');
+  assert.strictEqual(this.player.ads.state, 'content-playback');
+  this.player.paused = function() {
+    return false;
+  };
+  sinon.spy(this.player, 'pause');
+  this.player.trigger('contentupdate');
+  assert.strictEqual(this.player.ads.state, 'ads-ready?');
+  assert.ok(this.player.pause.calledOnce, 'player was paused');
+});
+
 // ----------------------------------
 // Event prefixing during ad playback
 // ----------------------------------
--- a/test/test.ads.js
+++ b/test/test.ads.js
@@ -729,6 +729,16 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
   assert.strictEqual(spy.getCall(0).args[0].type, 'readyforpreroll', 'readyforpreroll should have been triggered.');
 });
 
+QUnit.test('contentupdate in preroll? transitions to content-set', function(assert) {
+  assert.expect(2);
+
+  this.player.trigger('play');
+  this.player.trigger('adsready');
+  assert.strictEqual(this.player.ads.state, 'preroll?');
+  this.player.trigger('contentupdate');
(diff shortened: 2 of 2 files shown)


src/plugin.js
```diff
@@ -666,6 +670,10 @@ const contribAdsPlugin = function(options) {
           if (player.paused()) {
             this.state = 'content-set';
           } else {
+            // If we get here we missed the 'play' event, so pause the player so
+            // that content doesn't start playing immediately after the source change
+            player.pause();
```
ldayananda: I think we may want to pause the player if it's not paused yet and then go to the **same** state `content-set`. 

If I understand correctly, the reason this else block will go to `ads-ready?` is that there is an assumption that the `play` event has been missed. So once the player is paused, we should be in the same state that causes the if clause to change state to `content-set`

alex-barstow: Makes sense to me

ldayananda: From an offline conversation we came to the conclusion that the state transition to `ads-ready?` is needed still.

test/test.ads.js
```diff
@@ -729,6 +729,45 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
   assert.strictEqual(spy.getCall(0).args[0].type, 'readyforpreroll', 'readyforpreroll should have been triggered.');
 });
 
+QUnit.test('contentupdate in preroll? transitions to content-set', function(assert) {
+  assert.expect(2);
+
```
ldayananda: I believe all these tests should trigger `loadstart` before play as well, to make sure the player is setup correctly

test/test.ads.js
```diff
@@ -729,6 +729,45 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
...
+  assert.strictEqual(this.player.ads.state, 'preroll?');
+  this.player.trigger('contentupdate');
+  assert.strictEqual(this.player.ads.state, 'content-set');
+});
+
+QUnit.test('contentupdate in content-playback transitions to content-set if the player is paused', function(assert) {
+  assert.expect(2);
+
+  this.player.trigger('play');
+  this.player.trigger('adtimeout');
```
ldayananda: I would just do `this.players.ads.skipLinearAdMode()` here, or explicitly do `startLinearAdMode()` and  `endLinearAdMode()` right after.

test/test.ads.js
```diff
@@ -729,6 +729,45 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
...
+  };
+  this.player.trigger('contentupdate');
+  assert.strictEqual(this.player.ads.state, 'content-set');
+});
+
+QUnit.test('contentupdate in content-playback transitions to ads-ready? and pauses player if not already paused', function(assert) {
+  assert.expect(4);
+
+  this.player.trigger('play');
+  this.player.trigger('adtimeout');
```
ldayananda: same feedback for this line as the above test

src/plugin.js
```diff
@@ -435,6 +436,9 @@ const contribAdsPlugin = function(options) {
         },
         nopreroll() {
           this.state = 'content-playback';
+        },
+        contentupdate() {
+          this.state = 'content-set';
```
incompl: There is a loadstart/contentupdate for the initial source. Is there any guarantee that this isn't that event? Specifically, if the play / adsready came in before loadstart.

test/test.ads.js
```diff
@@ -729,6 +729,46 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
...
 
+QUnit.test('contentupdate in preroll? transitions to content-set', function(assert) {
+  assert.expect(2);
+
+  this.player.trigger('loadstart');
+  this.player.trigger('play');
+  this.player.trigger('adsready');
+  assert.strictEqual(this.player.ads.state, 'preroll?');
+  this.player.trigger('contentupdate');
+  assert.strictEqual(this.player.ads.state, 'content-set');
```
incompl: You should avoid tests that assert the state. Firstly, they will break very shortly. Secondly, it is preferable to test things that indicate that the player is functioning correctly. For example, making sure it is paused if it should be paused.

test/test.ads.js
```diff
@@ -729,6 +729,46 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
   assert.strictEqual(spy.getCall(0).args[0].type, 'readyforpreroll', 'readyforpreroll should have been triggered.');
 });
 
+QUnit.test('contentupdate in preroll? transitions to content-set', function(assert) {
+  assert.expect(2);
```
incompl: I recommend avoiding using `expect` unless your code has assertions in a loop. For simple procedural tests, the expect doesn't add value and requires maintenance if the test is updated.

test/test.ads.js
```diff
@@ -729,6 +729,46 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
...
+  this.player.trigger('loadstart');
+  this.player.ads.skipLinearAdMode();
+  assert.strictEqual(this.player.ads.state, 'content-playback');
+  this.player.paused = function() {
+    return false;
+  };
+  sinon.spy(this.player, 'pause');
+  this.player.trigger('contentupdate');
+  assert.strictEqual(this.player.ads.state, 'ads-ready?');
+  assert.ok(this.player.pause.calledOnce, 'player was paused');
```
incompl: The assertions on pause.calledOnce and _pausedOnContentUpdate are good, you can remove the assertions on the state. The test remains just as valuable and would survive the refactor.

test/test.ads.js
alex-barstow: If that's the case, I'll remove these first two tests in their entirety since they are just testing that certain state transitions occur under the proper conditions.

incompl: Makes sense, it does seem like the third test covers the functionality you've implemented here.

src/plugin.js
alex-barstow: Good point. I'm not sure there is any such guarantee. We could add a safeguard, something like:
```js
contentupdate() {
  // only change states if this is not the first loadstart/contentupdate
  if (player.ads._hasThereBeenALoadStartDuringPlayerLife) {
    this.state = 'content-set';
  }
}
```

incompl: You'll need to test it to make sure `_hasThereBeenALoadStartDuringPlayerLife` is set, it's also set on loadstart so the order of event handlers matters here.

alex-barstow: In hindsight, this check won't have any value for subsequent videos in a playlist since `_hasThereBeenALoadStartDuringPlayerLife` will be true for the life of the player after the first `loadstart`, as the name suggests :P

test/test.ads.js
```diff
@@ -729,6 +729,18 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
   assert.strictEqual(spy.getCall(0).args[0].type, 'readyforpreroll', 'readyforpreroll should have been triggered.');
 });
 
+QUnit.test('contentupdate in content-playback transitions to ads-ready? and pauses player if not already paused', function(assert) {
```
incompl: Minor: remove state reference from test name

test/test.ads.js
```diff
@@ -729,6 +729,18 @@ QUnit.test('adsready in content-playback triggers readyforpreroll', function(ass
...
+  this.player.trigger('loadstart');
+  this.player.ads.skipLinearAdMode();
+  this.player.paused = function() {
+    return false;
+  };
+  sinon.spy(this.player, 'pause');
+  this.player.trigger('contentupdate');
+  assert.ok(this.player.pause.calledOnce, 'player was paused');
+  assert.ok(this.player.ads._pausedOnContentupdate, '_pausedOnContentupdate is true');
```
incompl: Can we also test that the play happens?

src/plugin.js
```diff
@@ -666,6 +667,10 @@ const contribAdsPlugin = function(options) {
           if (player.paused()) {
             this.state = 'content-set';
           } else {
+            // If we get here we missed the 'play' event, so pause the player so
+            // that content doesn't start playing immediately after the source change
+            player.pause();
```
incompl: What about if contentupdate for the initial source happens during content-playback? This can happen with autoplay.

alex-barstow:
This fix has been moved into the [refactor](https://github.com/videojs/videojs-contrib-ads/pull/315). Closing.

PR #329, opened 2018-02-23
chore: fix changelog

incompl:
Removing bower because it messes up the changelog tooling and bower is going the way of the dinosaur anyhow.


incompl:
@ldayananda I believe not.

PR #337, opened 2018-02-28
feat: Add Play middleware to avoid calling play on tech

ldayananda:
Limitations: can't be used on mobile platforms and is only available for use on later versions of videojs.

- Implements a terminating play middleware.
- Removes special `adplaying` handling
- Reworks the "fast track" cases where there is no preroll ad to first advance to Preroll state and contentResuming mode
- Simplify logic to determine whether a manual play is required after resuming to content
- Further modularizes cancelContentPlay and removes 1ms timeout

Todos:
- [x] write integration tests for play middleware
- [x] finalize mocking for unit tests for play middleware
- [x] more negative test cases in unit tests


src/plugin.js
```diff
@@ -315,6 +325,7 @@ const contribAdsPlugin = function(options) {
   };
 
   player.ads._state = new BeforePreroll(player);
+  player.ads._state.init(player);
```
incompl: Ah interesting, was this causing an issue or was it just something you noticed?

src/states/BeforePreroll.js
```diff
@@ -30,7 +33,10 @@ export default class BeforePreroll extends ContentState {
     player.ads.debug('Received play event (BeforePreroll)');
 
     // Don't start content playback yet
-    cancelContentPlay(player);
+    if (!isMiddlewareMediatorSupported()) {
```
incompl: Maybe you already planned on doing this, but the next step is to move this logic into cancelContentPlay.js. That is, always `cancelContentPlay` on `play` if `!isMiddlewareMediatorSupported`. Then all the states have to do is set `_shouldBlockPlay`.

src/plugin.js
ldayananda: Mostly something I noticed, I wanted to set terminate at the earliest point possible in BeforePreroll but felt putting it in the constructor wasn't right

incompl: Yup makes total sense.

migration-guides/migrating-to-6.0.md
```diff
@@ -18,6 +18,7 @@ by integrations.
 * adTimeoutTimeout has been removed. It was not part of the documented interface, but make note if your integration inspected it.
 * There is no longer a snapshot object while checking for postrolls. Now a snapshot is only taken when a postroll ad break actually begins.
 * The `contentplayback` event (removed in [4.0.0](https://github.com/videojs/videojs-contrib-ads/blob/cc664517aa0d07398decc0aa5d41974330efc4e4/CHANGELOG.md#400), re-added as deprecated in [4.1.1](https://github.com/videojs/videojs-contrib-ads/blob/cc664517aa0d07398decc0aa5d41974330efc4e4/CHANGELOG.md#411)), has been removed. Use the `playing` event instead.
+* The `adplaying` behavior is an implementation detail and has changed in this update. The `adplaying` event is no longer guaranteed to happen once per ad break. It is not intended to be used to detect the beginning of an ad break. The `ads-pod-started` event should be used instead. The `ads-ad-started` event can be used to detect the start of an individual ad in an ad break. There will be multiple `ads-ad-started` events corresponding to each ad in the ad break.
```
incompl: 👍

src/cancelContentPlay.js
```diff
@@ -6,11 +8,19 @@ then signalling that we should play after the ad is done.
 */
 
 export default function cancelContentPlay(player) {
-  if (player.ads.cancelPlayTimeout) {
+  if (pm.isMiddlewareMediatorSupported()) {
+    // Don't use cancelContentPlay while playMiddleware is in use
+    return;
+  } else if (player.ads._shouldBlockPlay === false) {
+    // Don't block play if explictly set to not block
```
incompl: To make this comment more specific, maybe "Only block play if the ad plugin is in a state where play should be blocked. This currently means BeforePreroll and Preroll."

src/cancelContentPlay.js
```diff
@@ -6,11 +8,19 @@ then signalling that we should play after the ad is done.
...
+    return;
+  } else if (player.ads.cancelPlayTimeout) {
     // another cancellation is already in flight, so do nothing
     return;
   }
 
+  player.ads.debug('Using cancelContentPlay to block content playback');
```
incompl: 👍

src/playMiddleware.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+obj.playMiddleware = function(player) {
+  return {
+    setSource(srcObj, next) {
+      next(null, srcObj);
+    },
+    callPlay() {
+      // Block play calls while waiting for an ad
+      if (obj.isMiddlewareMediatorSupported() &&
```
incompl: Is the check to `isMiddlewareMediatorSupported` needed?

src/playMiddleware.js
```diff
@@ -0,0 +1,50 @@
...
+          player.ads._shouldBlockPlay === true) {
+        player.ads.debug('Using playMiddleware to block content playback');
+        return videojs.middleware.TERMINATOR;
+      }
+    },
+    play(terminated, value) {
+      if (terminated) {
+        player.ads.debug('Play event was terminated.');
+        player.trigger('play');
```
incompl: I would suggest a comment explaining why we trigger play after play was terminated.

src/plugin.js
```diff
@@ -9,6 +9,7 @@ import redispatch from './redispatch.js';
 import initializeContentupdate from './contentupdate.js';
 import adMacroReplacement from './macros.js';
 import cueTextTracks from './cueTextTracks.js';
+import pm from './playMiddleware.js';
```
incompl: I suggest spelling it out as `import playMiddleware from './playMiddleware.js';` to follow the project convention and make the code more readable.

src/playMiddleware.js
incompl: I ask because it's checked before setting up the middleware in plugin.js

src/plugin.js
```diff
@@ -120,6 +126,14 @@ const contribAdsPlugin = function(options) {
     player.ads._pausedOnContentupdate = false;
   });
 
+  // Keep track of whether a play event has happened
+  player.on('play', () => {
+    videojs.log('*$*', '_playRequested about to be set');
```
incompl: Some debug messages to remove still

src/redispatch.js
```diff
@@ -126,12 +137,18 @@ const handleLoadEvent = (player, event) => {
...
 // after each one. To be continued...
 const handlePlay = (player, event) => {
-  const resumingAfterNoPreroll = player.ads._cancelledPlay && !player.ads.isInAdMode();
+  // eslint-disable-next-line no-console
+  console.log('**** resumingAfterNoAd', player.ads.isResumingAfterNoPreroll());
 
   if (player.ads.inAdBreak()) {
     prefixEvent(player, 'ad', event);
-  } else if (player.ads.isContentResuming() || resumingAfterNoPreroll) {
+  } else if (player.ads.isContentResuming() ||
```
incompl: I suggest a comment for this line

src/states/ContentPlayback.js
```diff
@@ -7,16 +8,44 @@ import {ContentState, Midroll, Postroll} from '../states.js';
...
+   */
+  isResumingAfterNoPreroll() {
+    return this.resumingAfterNoPreroll;
+  }
+
+  /*
+   * Reset resumingAfterNoPreroll once the content actually starts playing
+   */
+  onPlaying(player) {
+    this.resumingAfterNoPreroll = false;
```
incompl: I believe we transition to Content Mode after a `playing` event, so would this be the second playing event? Not sure that resumingAfterNoPreroll needs to be part of ContentPlayback and that the ad states could handle it.

incompl: In that case, would it work to set `this.resumingAfterNoPreroll = false;` in `ContentPlayback.init`?

src/states/Preroll.js
```diff
@@ -26,8 +26,13 @@ export default class Preroll extends AdState {
       player.trigger('adtimeout');
     }, timeout);
 
+    if (!this.inAdBreak() && !this.isContentResuming()) {
+      player.ads._shouldBlockPlay = true;
```
incompl: In init you would never be in ad break or content resuming, so this check isn't needed.

src/states/abstract/State.js
```diff
@@ -86,6 +86,13 @@ export default class State {
     return false;
   }
 
+  /*
+   * Overriden by BeforePrerollState, PrerollState and ContentPlaybackState.
```
incompl: Don't need `State` in the names

src/playMiddleware.js
ldayananda: This is mostly defensive. Although it is checked before registering the the middleware, if in case that changes we want to make sure we don't use both the playMiddleware and cancelContentPlay at the same time.

incompl: I think in this case it doesn't have much defensive value but could lead to confusion about how the code works. That's subjective though so not a big deal.

src/states/Preroll.js
ldayananda: This is mostly defensive since the flag is used for both cancelContentPlay and playMiddleware and the timing of when it is set is important.

ldayananda: I would also like to avoid retesting it currently

src/states/abstract/State.js
ldayananda: Ah, I was following the convention elsewhere in this file

src/states/ContentPlayback.js
incompl: Ah, I think this further indicates an issue with how we're handling nopreroll. Normally we transition to ContentPlayback after playing, but in this case we don't. As a result, preroll logic is creeping into ContentPlayback.

What do you think of setting resumingAfterNoPreroll within Preroll, and transitioning to Preroll on nopreroll (heh) with resumingAfterNoPreroll and contentResuming both set to true?

src/states/abstract/State.js
incompl: It's probably my bad, I renamed them at one point during development. I think we should use the current file names.

src/states/ContentPlayback.js
ldayananda: so `Preroll -> Preroll`? 

Honestly I think this would be a big change and wouldn't fit in the context of this PR. I think making resumingAfterNoPreroll "first-class" should be done separately, and really we shouldn't have the special resumingAfterNoPreroll flag and instead go into contentResuming until `playing`.

incompl: I agree with avoiding the special `resumingAfterNoPreroll` flag and staying in Preroll until `playing`. The state transition would be `BeforePreroll -> Preroll` if `nopreroll` happened earlier. If `nopreroll` happened during `Preroll` it would stay in that state and set `contentResuming` to `true`.

The reason I think it fits into this PR is that you're creating technical debt (the preroll handling in ContentPlayback) to work around the problem. Rather than introduce new technical debt with the intent to remove it later, it's preferable to not introduce it at all.

I'm 100% down to assist in making this change.

src/states/ContentPlayback.js
```diff
@@ -1,4 +1,5 @@
 import {ContentState, Midroll, Postroll} from '../states.js';
+// import pm from '../playMiddleware.js';
```
ldayananda: Remove!

src/states/ContentPlayback.js
```diff
@@ -7,16 +8,36 @@ import {ContentState, Midroll, Postroll} from '../states.js';
  */
 export default class ContentPlayback extends ContentState {
 
-  init(player) {
-    // Play the content if cancelContentPlay happened or we paused on 'contentupdate'
+  init(player, resumingAfterNoPreroll) {
```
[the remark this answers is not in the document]
ldayananda: Created https://github.com/videojs/videojs-contrib-ads/issues/355 for this

test/unit/test.playMiddleware.js
```diff
@@ -0,0 +1,150 @@
...
+    this.sandbox.stub(videojs, 'browser').get(() => {
+      return {
+        IS_ANDROID: false,
+        IS_IOS: false
+      };
+    });
+  }
+}));
+
+QUnit.skip('isMiddlewareMediatorSupported is true if middleware mediators exist on desktop', function(assert) {
```
incompl: Should these be skipped?

test/unit/test.redispatch.js
```diff
@@ -75,9 +89,44 @@ QUnit.test('playing event in different ad states', function(assert) {
...
+QUnit.test('play events in different states', function(assert) {
+  this.player.ads.inAdBreak = () => false;
   this.player.ads.isInAdMode = () => true;
+  this.player.ads.isContentResuming = () => true;
+  assert.equal(this.redispatch('play'), 'contentplay',
+    'should be contentplay when resuming after nopreroll');
+
+  this.player.ads.inAdBreak = () => false;
+  this.player.ads.isInAdMode = () => false;
+  this.player.ads.isContentResuming = () => true;
```
incompl: content resuming is part of ad mode, so this is not a valid scenario

test/unit/test.redispatch.js
```diff
@@ -33,6 +33,20 @@ QUnit.module('Redispatch', {
 
         stitchedAds() {
           return false;
+        },
+
+        isResumingAfterNoPreroll() {
+          return false;
+        },
+
+        _state: {
```
incompl: Why is this needed now?

src/playMiddleware.js
ldayananda: So here's my thinking behind the "defensive code" reasoning:
- once a middleware is registered, it will always run(since we're using the MIME type agnostic '*')
- there's a number of video.js versions that don't have Middleware Mediator support (since it was added relatively recently)
- since the middleware is registered in plugin.js, I'm worried that the reason why we check  `isMiddlewareMediatorSupported` would be lost over time and we'll wind up doing something that causes a runtime exception.

Maybe I'm just worrying about nothing though.

src/plugin.js
```diff
@@ -163,6 +180,10 @@ const contribAdsPlugin = function(options) {
     // Are we after startLinearAdMode and before endLinearAdMode?
     _inLinearAdMode: false,
 
+    //
```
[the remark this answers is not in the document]
ldayananda: Oops, meant to fill this out

src/redispatch.js
```diff
@@ -127,18 +126,22 @@ const handleLoadEvent = (player, event) => {
...
+// video.js after version 6.7.1.
 const handlePlay = (player, event) => {
-  const resumingAfterNoPreroll = player.ads._cancelledPlay && !player.ads.isInAdMode();
-
   if (player.ads.inAdBreak()) {
     prefixEvent(player, 'ad', event);
-  } else if (player.ads.isContentResuming() || resumingAfterNoPreroll) {
+  // If a play event has already happened and a preroll did not play,
+  // the next play event that is not user-initiated should be
+  // prefixed as though we are resuming to content.
```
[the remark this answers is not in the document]
ldayananda: Whoops, this was a comment I wrote before changing the nopreroll flow. Good catch!

src/states/Preroll.js
```diff
@@ -21,10 +20,19 @@ export default class Preroll extends AdState {
...
     player.addClass('vjs-ad-loading');
 
+    // If adserror, adscanceled, nopreroll or skipLinearAdMode already
+    // ocurred, resume to content immediately
+    if (shouldResumeToContent || player.ads.nopreroll_) {
+      return this.resumeAfterNoPreroll(player);
+    }
+
+    // Continue to block content playback until an ad break starts
+    player.ads._shouldBlockPlay = true;
```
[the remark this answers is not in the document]
ldayananda: I believe no 🤔 though I haven't tested whether I can remove this safely yet.

src/states/Preroll.js
```diff
@@ -225,6 +233,24 @@ export default class Preroll extends AdState {
...
+    this.contentResuming = true;
+    player.ads._shouldBlockPlay = false;
+
+    player.ads.debug('paused?', player.paused(), player.ads._playRequested || player.ads._pausedOnContentupdate);
+    // Play the content if we had requested play or we paused on 'contentupdate'
+    // and we haven't played yet. This happens if there was no preroll or if it
+    // errored, timed out, etc. Otherwise snapshot restore would play.
+    if (player.paused() &&
+        (player.ads._playRequested || player.ads._pausedOnContentupdate)) {
+      player.ads.debug('Playing again since the player is paused');
```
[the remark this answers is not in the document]
ldayananda: I thought it might be useful if we're debugging issues with resuming after nopreroll type situations. Thoughts?

test/integration/test.playMiddleware.js
```diff
@@ -0,0 +1,64 @@
...
+  videojs.log('second loadstart');
+  this.player.trigger('loadstart');
+  assert.strictEqual(this.player.ads._playRequested, false,
+    '_playRequested reset');
+  assert.strictEqual(spy.callCount, 1,
+    'contentchanged once');
+
+  this.clock.tick(1);
+
+  const testAssert = function(player, clock) {
```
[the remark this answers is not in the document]
ldayananda: The tests aren't 100% quite yet

test/unit/test.playMiddleware.js
ldayananda: These tests still need to be updated to mock out video.js completely

test/unit/test.redispatch.js
ldayananda: 🤔 I might be able to remove this now actually

src/playMiddleware.js
incompl: Once middleware is registered it will always run, and conditions for it being registered will never change, so it won't need to be checked again.

The video.js versions that don't have mediator support are handled within the check before registering.

The last concern is definitely a valid one, but I think it is best addressed by commenting the code to add the reason. Basically it sounds like you're writing a workaround for a hypothetical bug, but really, you're making it harder to identify the hypothetical bug when if and when it happens.

src/states/Preroll.js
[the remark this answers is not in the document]
incompl: Yeah I'm not sure, my thinking is that it's pretty clear when there is a problem because the player will either play or not play, so you'd see it visually. Beyond that you'd need to go into the debugger anyway.

src/playMiddleware.js
ldayananda: Maybe the last concern can be addressed by doing two things:
- adding more comments to explain why we check for support first in plugin.js
- check if videojs.middleware exists before trying to return the middleware.TERMINATOR here, to avoid a runtime exception breaking the whole player. 

Also, this check is also done in `initCancelContentPlay`, do you feel the same about that?

incompl: "adding more comments to explain why we check for support first in plugin.js" 👍 

"check if videojs.middleware exists before trying to return the middleware.TERMINATOR" That doesn't seem much different. If middleware isn't supported, we would not have gotten that far.

"Also, this check is also done in initCancelContentPlay, do you feel the same about that?" I do, yeah. It's performing a check for an unrelated feature. The logic should be hoisted to the first place that can handle it and form a bottleneck rather than spreading checks throughout various modules.

src/states/Preroll.js
```diff
@@ -144,11 +135,12 @@ export default class Preroll extends AdState {
     // if there was an error.
     if (this.inAdBreak()) {
       player.ads.endLinearAdMode();
-    }
 
-    this.afterLoadStart(() => {
-      this.transitionTo(ContentPlayback);
-    });
+    } else {
```
ldayananda: Does this change seem right @incompl? This way both endLinearAdMode and resumeAfterNoPreroll will wait for `playing` to transitionTo ContentPlayback

test/integration/test.ads.js
```diff
@@ -214,7 +154,7 @@ QUnit.test('changing the src triggers "contentupdate"', function(assert) {
   this.player.on('contentupdate', spy);
 
   // set src and trigger synthetic 'loadstart'
-  this.player.src('http://media.w3.org/2010/05/sintel/trailer.mp4');
+  this.player.currentSrc = () => 'AH HA!!! I AM NOT A REAL SOURCE';
```
ldayananda: @incompl do we want to change this back?

test/integration/test.cancelContentPlay.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+// Run custom hooks before sharedModuleHooks, as videojs must be
+// modified before seting up the player and videojs-contrib-ads
+QUnit.module('Cancel Content Play', {
+  beforeEach: _.flow(fakeVideojs, sharedHooks.beforeEach),
+  afterEach: _.flow(restoreVideojs, sharedHooks.afterEach)
+});
+
+QUnit.test('pauses to wait for prerolls when the plugin loads BEFORE play', function(assert) {
```
ldayananda: These tests were moved from `test.ads.js`

src/states/Preroll.js
incompl: adserror is deprecated and I think we should focus on removing reliance on it rather than fixing it.

test/integration/test.ads.js
incompl: It's kind of nice that this demonstrates that a source isn't really being loaded, but we could also make the string less annoying.

src/playMiddleware.js
```diff
@@ -0,0 +1,60 @@
+import videojs from 'video.js';
+
+const obj = {};
+let videojsReference = videojs;
```
incompl: A comment for this line would be helpful. It's not obvious why this is necessary.

src/plugin.js
```diff
@@ -66,6 +69,19 @@ const contribAdsPlugin = function(options) {
   // Set up redispatching of player events
   player.on(videoEvents, redispatch);
 
+  // Set up features to block content playback while waiting for ads.
+  // Play middleware is only supported on later versions of video.js
+  // and on desktop currently(as the user-gesture requirement on mobile
+  // will disallow calling play once play blocking is lifted)
+  // The middleware must also be registered outside of the plugin,
+  // to avoid a middleware factory being created for each player
```
incompl: Good comment

src/plugin.js
```diff
@@ -66,6 +69,19 @@ const contribAdsPlugin = function(options) {
...
+  if (isMiddlewareMediatorSupported() && settings.debug) {
+    videojs.log('ADS:', 'Play middleware has been registered with videojs');
```
incompl: It would be a little tidier to move this to where middleware is actually registered.

src/states/Preroll.js
```diff
@@ -21,10 +20,16 @@ export default class Preroll extends AdState {
...
    * happen here, not in a constructor.
    */
-  init(player, adsReady) {
+  init(player, adsReady, shouldResumeToContent) {
     // Loading spinner from now until ad start or end of ad break.
     player.addClass('vjs-ad-loading');
 
+    // If adserror, adscanceled, nopreroll or skipLinearAdMode already
+    // ocurred, resume to content immediately
+    if (shouldResumeToContent || player.ads.nopreroll_) {
```
incompl: I realize it may be late in the game to mess with this, but is the `player.ads.nopreroll_` part still necessary? The `shouldResumeToContent` stuff seems pretty airtight.

incompl:
LGTM, just some minor comments

src/states/Preroll.js
ldayananda: The main reason for it is that we set `player.ads.nopreroll_` globally after listening for the `nopreroll` event in plugin.js. My logic was that it's still possible for there to be a nopreroll event  roughly at the same time as play/one of the other negative scenarios and so we should check just in case. If we go down the route of removing `nopreroll_` from the global state and making it part of the BeforePreroll & Preroll states only, I think we would be safe to remove it then

src/states/Preroll.js
ldayananda: Cool, I think we agree that this change makes sense then?

incompl: 👍

src/states/Preroll.js
incompl: Yeah that sounds like the right direction to go in. Don't need to do it now.

bcsdasneogi:
QA : LGTM

PR #350, opened 2018-03-15
fix: Couble ended event bug

incompl:
Because we no longer have a delay before firing the ended event, we would get a case where snapshot restore would fire an additional ended event. This PR adds a callback for adBreak.done and restorePlayerSnapshot so we can wait until the snapshot restore is done before firing ended. To make this reliable, this PR also simplifies redispatch for ended and fires ended only in AdsDone.init.


README.md
```diff
@@ -111,6 +110,7 @@ Deprecated events:
 
 * `contentupdate` (EVENT) -- Replaced by `contentchanged`, which is more reliable.
 * `adscanceled` (EVENT) -- Intended to cancel all ads, it was never fully implemented. Instead, use `nopreroll` and `nopostroll`.
+* `adserror` (EVENT) -- This event skips prerolls when seen before a preroll ad break. It skips postrolls if called after contentended and before a postroll ad break. It ends linear ad mode if seen during an ad break. It can be replaced using `skipLinearAdMode` and `endLinearAdMode`.
```
ldayananda: I don't think adserror was actually deprecated

examples/basic-ad-plugin/example-integration.js
```diff
@@ -104,6 +104,8 @@
       if (!state.postrollPlayed && playPostroll) {
         state.postrollPlayed = true;
         playAd();
+      } else {
+        player.trigger('nopostroll');
```
ldayananda: 👍

src/adBreak.js
```diff
@@ -61,14 +64,19 @@ function end(player) {
...
+    snapshot.restorePlayerSnapshot(player, player.ads.snapshot, callback);
   }
 
   // Reset the volume to pre-ad levels
   if (player.ads.shouldPlayContentBehindAd(player)) {
     player.volume(player.ads.preAdVolume_);
   }
 
+  // We're done because we're not restoring a snapshot
+  if (player.ads.shouldPlayContentBehindAd(player)) {
```
ldayananda: Why not include this in the previous if?

src/redispatch.js
```diff
@@ -67,15 +67,9 @@ const handlePlaying = (player, event) => {
...
 
-    // The true ended event fired either after the postroll
-    // or because there was no postroll.
-    if (player.ads.isContentResuming()) {
-      return;
-    }
-
-    // Prefix ended due to ad ending.
+  // Prefix ended events in ad mode
+  if (player.ads.isInAdMode()) {
```
ldayananda: 💯

src/snapshot.js
```diff
@@ -188,6 +193,10 @@ export function restorePlayerSnapshot(player, snapshotObject) {
   // restoration is required
 
   if (player.ads.videoElementRecycled()) {
+    // Snapshot restore will result in an adended event. That's when
+    // we're really done.
+    player.one('adended', callback);
```
ldayananda: are there going to be multiple adended events in some scenarios? One from here and one from redispatch?

src/states/Postroll.js
```diff
@@ -37,11 +37,9 @@ export default class Postroll extends AdState {
 
     // No postroll, ads are done
     } else {
-      player.setTimeout(() => {
-        player.ads.debug('Triggered ended event (no postroll)');
-        this.contentResuming = true;
-        player.trigger('ended');
-      }, 1);
+      player.ads.debug('Triggered ended event (no postroll)');
```
ldayananda: seems like this comment should go now?

README.md
incompl: I'm deprecating it in this PR.

src/adBreak.js
incompl: Doing that

src/snapshot.js
incompl: yeah... normally you'd want the one that happens during content restoration to be `contentended` but that name was already used for something else. I figured calling it adended would get it out of the way. Ideally `contentended` would become `readyforpostroll` and then we could use `contentended` for this to match everything else. Don't think this is the time to do that though.

src/states/Postroll.js
incompl: Removing it

src/snapshot.js
incompl: We could also make a new event like `restoreended` if we want. IDK if that's any better.

src/adBreak.js
```diff
@@ -61,12 +64,13 @@ function end(player) {
     player.addClass('vjs-live');
   }
   if (!player.ads.shouldPlayContentBehindAd(player)) {
-    snapshot.restorePlayerSnapshot(player, player.ads.snapshot);
+    snapshot.restorePlayerSnapshot(player, player.ads.snapshot, callback);
   }
 
   // Reset the volume to pre-ad levels
   if (player.ads.shouldPlayContentBehindAd(player)) {
```
incompl: Make this an else

src/redispatch.js
```diff
@@ -67,15 +67,9 @@ const handlePlaying = (player, event) => {
 // * No ended event before postroll
 // * A single ended event after postroll
 const handleEnded = (player, event) => {
-  if (player.ads.isInAdMode()) {
 
-    // The true ended event fired either after the postroll
-    // or because there was no postroll.
-    if (player.ads.isContentResuming()) {
```
incompl: Per code review: `trigger('_snapshotended')`

incompl: And cancel the event

README.md
```diff
@@ -534,10 +534,10 @@ A short list of features, fixes and changes for each release is available in [CH
...
-* `contentended` has a confusing name: real `ended` events are later sent, and that is when content should be considered ended. The `content` prefix is used for events when content is resuming after an ad. A better name would be `readyforpostroll`. That would make it clearer to implementations that the correct response would be to either play a postroll or send the `nopostroll` event.
+* `contentended` has a confusing name: real `ended` events are later sent, and that is when content should be considered ended. The `content` prefix is used for events when content is resuming after an ad. A better name would be `readyforpostroll`. That would make it clearer to implementations that the correct response would be to either play a postroll or send the `nopostroll` event. `resumeended`, which is a workaround for this issue, could then become `contentended`, which would mirror all the other redispatched events.
```
ldayananda: 💯

migration-guides/migrating-to-6.0.md
```diff
@@ -25,7 +25,7 @@ Deprecated interfaces will be removed in a future major version update.
...
 * `adscanceled` is now deprecated. Instead, use `nopreroll` and `nopostroll`. `adscanceled` was initially intended to function similarly to calling both `nopreroll` and `nopostroll` but it was never fully implemented.
-* `adserror` is now deprecated. Currently this event will skip prerolls when seen before a preroll ad break, skip postrolls if called after contentended and before a postroll ad break, and end linear ad mode if seen during an ad break. It is better for the ad integration to do these things explicitely with `skipLinearAdMode` and `endLinearAdMode`. In the future, this event will not have any special behavior.
+* `adserror` is now deprecated. Currently this event will skip prerolls when seen before a preroll ad break, skip postrolls if called after contentended and before a postroll ad break, and end linear ad mode if seen during an ad break. It is better for the ad integration to do these things explicitely with `skipLinearAdMode` and `endLinearAdMode`. In the future, this event will not have any special behavior. The event isn't being "removed" in any sense because contrib-ads does not trigger this event, so integrations may continue to use it for other purposes.
```
ldayananda: Typo `explicitely`. Also rephrase to say `skipLinearAdMode` and `endLinearAdMode` are the supported paths rather than being better

src/redispatch.js
```diff
@@ -67,12 +67,26 @@ const handlePlaying = (player, event) => {
...
+    // Cancel ended events during content resuming. Normally we would
+    // prefix them, but `contentended` has a special meaning. In the
+    // future we'd like to rename the existing `contentended` to
+    // `readyforpostroll`, then we could remove the special `resumeended`
+    // and do a conventional content prefix here.
+    if (player.ads.isContentResuming()) {
+      cancelEvent(player, event);
+
+      // Important: do not use this event outside of videojs-contrib-ads.
+      // It will be removed and your code will break.
```
ldayananda: 😄

migration-guides/migrating-to-6.0.md
incompl: Well, it is better, so I'll say both

incompl: The supported part is clear from the context, maybe "It is more declarative for the ad integration to do these things explicitly with `skipLinearAdMode` and `endLinearAdMode`. "

bcsdasneogi:
QA : LGTM

PR #365, opened 2018-04-19
Play middleware: only trigger play event when our play middleware terminates

ldayananda:
This ensures that the code in playMiddleware.play does not run if another middleware terminates.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -193,7 +193,7 @@ const contribAdsPlugin = function(options) {
 
     // Should we block calls to play on the content player?
     _shouldBlockPlay: false,
-    // Was play blocked by the plugin?
+    // Was play blocked by the plugin's playMiddleware feature?
     _playBlocked: false,
     // Tracks whether play has been requested for this source,
     // either by the play method or user interaction
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -193,6 +193,8 @@ const contribAdsPlugin = function(options) {
 
     // Should we block calls to play on the content player?
     _shouldBlockPlay: false,
+    // Was play blocked by the plugin?
+    _playBlocked: false,
     // Tracks whether play has been requested for this source,
     // either by the play method or user interaction
     _playRequested: false,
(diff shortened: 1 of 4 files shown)


src/plugin.js
```diff
@@ -193,6 +193,8 @@ const contribAdsPlugin = function(options) {
 
     // Should we block calls to play on the content player?
     _shouldBlockPlay: false,
+    // Was play blocked by the plugin?
```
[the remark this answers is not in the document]
incompl: So that someone reading it knows what feature it's related to

test/integration/test.playMiddleware.js
```diff
@@ -142,4 +142,63 @@ QUnit.test("playMiddleware doesn\'t block play in content playback", function(as
...
+  localPlayer.on('play', playSpy);
+
+  // Wait for the middleware to run
+  localPlayer.setTimeout(() => {
+    assert.strictEqual(localPlayer.ads._playBlocked, false,
+      'play should not have been blocked');
+    assert.strictEqual(playSpy.callCount, 0,
+      'play event should not be triggered');
+    done();
+  }, 200);
```
incompl: Could this timeout replaced by an event handler?

ldayananda: I don't believe so, as if the play is blocked by another middleware we neither advance in the state machine nor do we start playback. That said, if you have another idea I'm all ears 😄

incompl: Would it work if it was 0 or 1 ms instead of 200? 200 makes it seem like it's waiting for _something_

ldayananda: Looks like 1ms was actually enough!

incompl: Great, that should make it a little less mysterious

ldayananda:
@incompl updated based on comments

incompl:
LGTM

rtezera1:
QA: Pass!

PR #378, opened 2018-05-04
fix: Update link with initialization info

incompl:


src/plugin.js
```diff
@@ -88,7 +88,7 @@ const contribAdsPlugin = function(options) {
         'videojs-contrib-ads was initialized too late. It must be initialized ' +
         'immediately after video.js in the same tick. As a result, some ads will not ' +
         'play and some media events will be incorrect. For more information, see ' +
-        'https://github.com/videojs/videojs-contrib-ads#important-note-about-initialization');
+        'http://videojs.github.io/videojs-contrib-ads/integrator/getting-started.html');
```
rtezera1: Checked why you chose `http` and not `https`. It seems that stylesheet of the page gets blocked because that's in http 😢

incompl: Oh weird. That's fixable, I'll make a separate PR

incompl: https://github.com/videojs/videojs-contrib-ads/pull/379

PR #380, opened 2018-05-08
IMA content flash after preroll  - preload none (onload)

pagarwal123:
When `startLinearAdMode()` is invoked, ad break starts where `removeNativePoster()` removes the poster attribute. When the ad break ends, `endLinearAdMode()` is invoked and adBreak.end(player) is called. Here, it restores the snapshot [restorePlayerSnapshot](https://github.com/videojs/videojs-contrib-ads/blob/c1f23c74f35702314b82e4d1468d29efad08dcdc/src/adBreak.js#L69) and adds the poster attribute [here](https://github.com/videojs/videojs-contrib-ads/blob/c1f23c74f35702314b82e4d1468d29efad08dcdc/src/snapshot.js#L182). Also, in videojs-ima3, it listens to the `playing` event and once again removes the poster attribute.
What happens in case of `preload:none` is, before it listens to `playing` and removes the poster attribute; there is already a flash of content(poster) as the poster attribute was added while restoring the snapshot.
Proposal: Do not add the `poster` attribute back when the ad break ends (as we are again removing it in videojs-ima3 once `playing` is triggered)

### DIFF
--- a/docs/integrator/snapshot.md
+++ b/docs/integrator/snapshot.md
@@ -4,6 +4,5 @@ The snapshot feature records the player state when an ad break begins and restor
 
 * Content source
 * Current time
-* Poster image
 * Style attribute
 * Text tracks
--- a/src/adBreak.js
+++ b/src/adBreak.js
@@ -36,8 +36,7 @@ function start(player) {
   }
 
   // This removes the native poster so the ads don't show the content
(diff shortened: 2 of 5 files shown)


rtezera1:
QA: PASS

PR #400, opened 2018-06-14
Play button sync and autoplay

incompl:


mysuf:
Nobody answered to your question so I will do. This fixes the issue. I can confirm that only for my integration so thats why I am replying here, but videojs-ima had same bug behaviour like my plugin so I guess that this fix will work also for videojs-ima. 

I also asked a question addressed to you: When contrib-ads now catches auto`play()` promise error (Chrome), it is one step from catching `pause()` too (which will probably called immediately after `play()`) and finally preventing triggering readyforpreroll. Is that possible? It seems that all of this heads to can-play own implementation..

incompl:
@mysuf Thanks for checking, we'll release it in the next contrib-ads release.

The idea about preventing readyforpreroll if play was blocked is on the right track. Maybe the best way to do this would be to expose a flag if autoplay was blocked. Then the ad plugin could decide not to trigger `adsready` (which would mean `readyforpreroll` would never happen) or skip the ad break if it's after `readyforpreroll`.

mysuf:
Yea, but both solutions should be considered as workarounds. I think that contrib-ads's role is actually middleware between player and ad plugin. `adsready` only tells to this middleware that ad plugin is prepared to play ad while blocked autoplay is state of player, which should be somehow handled.

I am asking if is it possible to catch blocked autoplay across browsers and trigger `readyforpreroll` only if video pass to true **playing** state. By contrib-ads.

incompl:
I'm not sure if I'd consider it a workaround. If we don't send `readyforpreroll` it prevents the ad plugin from being able to show a click-to-play ad, for example. I am going to think about this more.

mysuf:
Yea, but autoplay and click-to-play ad exludes each other.

rtezera1:
LGTM

pagarwal123:
QA-LGTM

PR #401, opened 2018-06-22
Chore/generator v6

BrandonOCasey:
## Actual Changes
* update the generator (which drops ie8-10 and older andriod browsers)
* Copy pasted the scss to css so that we can use postcss instead of scss
* Changed paths that point to `/base` to just point to `/` since karma now hosts a static file server

## Notes
* I noticed that manual tests on `localhost:9999/test` fail, presumable due to the recent autoplay changes
* I saw that rollup is pinned, seems to be due to circular dependency issues. I couldn't find a good way to fix it so I kept it pinned.

### Total gzip byte change
`-135`

### new
* `.min.js.gz`: 6973
* `.css.gz`: 384

### old
* `.min.js.gz`: 7075
* `.css.gz`: 417


gkatsev:
I think the `test/index.html` page has been broken for a while and probably should just be removed at this point. It's already recommended to use karma (https://github.com/videojs/videojs-contrib-ads/blob/master/docs/developer/getting-started.md#in-browser) and the new generator probably makes it easier too.

incompl:
Any plans for this PR? I see CI fails

BrandonOCasey:
Well tests pass locally, I can seem to figure out why one of the npm scripts seems to be failing, its very strange

incompl:
Looks like some kind of rollup issue in CI?

BrandonOCasey:
Everything appears to be fixed now.

PR #407, opened 2018-07-17
Handle dispose

incompl:

PR #412, opened 2018-07-24
restore sources instead of once source

mysuf:


mysuf:
I dont understand that particular test. You override player's `src()` method so how could be `tech_.src(source)` called when its calling was part of default `src()` method? I think that it ends by assigning `updatedSrc = [source]` (because this PR restores array instead of Object). And Array is not equal to expected Object...

mysuf:
With this modification, it just works and I believe it is the same (because `currentSrc()` and `currentType()` just returns properties from `currentSource()`:

```
// changing the src attribute to a URL that AdBlocker is intercepting
// doesn't update currentSrc, so when restoring the snapshot we
// should check for src attribute modifications as well
QUnit.test('checks for a src attribute change that isn\'t reflected in currentSrc', function(assert) {
  var updatedSrc;

  this.player.currentSource = function() {
    return {src: 'content.webm', type: 'video/webm'};
  };

  this.player.trigger('adsready');
  this.player.trigger('play');
  this.player.ads.startLinearAdMode();

  // `src` gets called internally to set the source back to its original
  // value when the player snapshot is restored when `endLinearAdMode`
  // is called.
  this.player.tech_.src = function(source) {
    if (source === undefined) {
      return 'ad.webm';
    }
    updatedSrc = source;
  };

  this.player.src = function(source) {
    if (source === undefined) {
      return 'ad.webm';
    }
    updatedSrc = source;
  };


  this.player.ads.endLinearAdMode();
  this.player.trigger('playing');
  assert.deepEqual(updatedSrc, [this.player.currentSource()], 'restored src attribute');
});
```

incompl:
Hmm, it looks like the test needs to mock `currentSources` instead of `currentSrc`, does that fix it? I'm not sure why the test has two places that it sets `updatedSrc` so maybe it could use cleanup.

mysuf:
1 test failed on something else on of nodes. Dunno why..

mysuf:
It cannot reach updatedSrc in tech_.src. I think that it should contain only getter `return 'ad.webm';`

gkatsev:
Seems fine to me, the only thing I'd be slightly concerned about, though, maybe it's not worth being concerned over it, is that if techs or source handlers are added during the ad, we may end up selecting a different source than what was originally selected.

mysuf:
Yep, it is not absolutely bulletproof. But adding some hooks during ad playback sounds error prone to me at all..

incompl:
I reran the build job that failed and it passed so I think it was a fluke.

I think selecting a different rendition after ads doesn't seem unreasonable, it may even be better in some cases.

LGTM

incompl:
I am currently testing stuff for a release of https://github.com/videojs/videojs-contrib-ads/pull/407, in order to not restart testing I'll do this as a second release after. Sorry for the delay.

PR #415, opened 2018-08-01
feat: More complete support for stitched ad scenarios.

misteroneill:
This PR is a re-implementation of #284.

This is a WIP still in that tests need fleshing out for these scenarios.


CONTRIBUTING.md
```diff
@@ -1,30 +1,3 @@
...
-
-- In all available and supported browsers: `npm test`
-- In a specific browser: `npm run test:chrome`, `npm run test:firefox`, etc.
-- While development server is running (`npm start`), navigate to [`http://localhost:9999/test/`][local]
-
-
-[karma]: http://karma-runner.github.io/
-[local]: http://localhost:9999/test/
-[conventions]: https://github.com/videojs/generator-videojs-plugin/blob/master/docs/conventions.md
+See: [docs/developer/index.md](docs/developer/index.md)
```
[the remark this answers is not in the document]
incompl: Awesome, I appreciate this

scripts/modules.rollup.config.js
```diff
@@ -9,8 +9,8 @@ import babel from 'rollup-plugin-babel';
 import json from 'rollup-plugin-json';
 
 export default {
-  moduleName: 'videojsContribAds',
-  entry: 'src/plugin.js',
+  name: 'videojsContribAds',
+  input: 'src/plugin.js',
```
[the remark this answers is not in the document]
incompl: Thanks!

src/ads.js
```diff
@@ -86,10 +86,17 @@ export default function getAds(player) {
       player.ads._state.skipLinearAdMode();
     },
 
+    // With no arguments, returns a boolean value indicating whether or not
+    // contrib-ads is set to treat ads as stitched with content in a single
+    // stream.
+    // With an argument, turns stitched ads mode on or off. This is best used
+    // immediately preceding a source change where the state would change to
+    // or from stitched ads.
```
incompl: Is it the same for the initial source, that it should be called before a source is set?

src/ads.js
```diff
@@ -130,12 +137,18 @@ export default function getAds(player) {
...
-    shouldPlayContentBehindAd(somePlayer) {
+    shouldPlayContentBehindAd(somePlayer = player) {
       return !videojs.browser.IS_IOS &&
              !videojs.browser.IS_ANDROID &&
              somePlayer.duration() === Infinity;
     },
 
+    // Return true if the ads plugin should save and restore snapshots of the
+    // player state when moving into and out of ad mode.
+    shouldTakeSnapshots(somePlayer = player) {
```
incompl: Nice

src/plugin.js
```diff
@@ -172,11 +172,11 @@ const contribAdsPlugin = function(options) {
   player.ads = getAds(player);
 
   player.ads.settings = settings;
+  player.ads.stitchedAds(settings.stitchedAds);
```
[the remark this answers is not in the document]
incompl: Makes sense, might be a good comment to add

src/ads.js
misteroneill: Yeah. There's a change in `plugin.js` that moves the setting of this to before the initial state is created. Otherwise, contrib-ads starts with `stitchedAds` being `false`.

src/states/BeforePreroll.js
```diff
@@ -23,6 +23,13 @@ export default class BeforePreroll extends ContentState {
     this.adsReady = false;
     this.shouldResumeToContent = false;
 
+    // Also, when ads are stitched into the content, content playback should
+    // not be blocked and we should move immediately into the Preroll state.
+    if (player.ads.stitchedAds()) {
+      player.ads._shouldBlockPlay = false;
+      return this.transitionTo(Preroll, this.adsReady, this.shouldResumeToContent);
```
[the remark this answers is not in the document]
misteroneill: Sounds good, I'll try it.

src/states/Preroll.js
```diff
@@ -21,28 +21,33 @@ export default class Preroll extends AdState {
...
    */
   init(player, adsReady, shouldResumeToContent) {
-    this.waitingForAdBreak = true;
 
-    // Loading spinner from now until ad start or end of ad break.
-    player.addClass('vjs-ad-loading');
+    // With stitched ads, we won't need to wait for a preroll and we can
+    // allow the integration to call startLinearAdMode whenever it's ready.
+    if (!player.ads.stitchedAds()) {
+      this.waitingForAdBreak = true;
```
incompl: Technically we're still waiting for ad break until startLinearAdMode, even though with stitched ads this will be a short time

incompl: I can see skipping the loading spinner, I wonder if nopreroll should be handled?

misteroneill: In our SSAI plugin, at least, we do fire it if there's no preroll stitched in. Makes sense to skip the spinner and preroll timeout only probably. I was thinking we might want the `resumeAfterNoPreroll` below.

incompl: 👍

misteroneill:
Note to self, tests to add:

- Prerolls do not block play.
- No `vjs-ad-loading` class is added for any roll.
- No timeout timers are queued for any roll.

src/ads.js
```diff
@@ -86,11 +86,20 @@ export default function getAds(player) {
...
+    // stream. With arguments, treated as a setter, but this behavior is
+    // deprecated.
     stitchedAds(arg) {
       if (arg !== undefined) {
-        this._stitchedAds = !!arg;
+        videojs.log.warn('Using player.ads.stitchedAds() as a setter is deprecated.');
+
+        // Keep the private property and the settings in sync. When this
+        // setter is removed, we can probably stop using the private property.
+        this._stitchedAds = this.settings.stitchedAds = !!arg;
```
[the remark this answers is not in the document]
misteroneill: We don't, good point. Removing.

examples/stitched-ad-plugin/plugin.js
```diff
@@ -0,0 +1,72 @@
...
+
+      // We need to tell contrib-ads that content has resumed because it
+      // otherwise expects a "play" event.
+      player.trigger('contentresumed');
+    });
+  };
+
+  var simulateStartOfAd = function(time) {
+    return waitForCurrentTime(time, function() {
+      player.on(['timeupdate', 'adtimeupdate'], simulateEndOfAd(time + 5));
```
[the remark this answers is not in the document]
misteroneill: That's probably correct. Copy/paste is the explanation, I'm sure.

misteroneill:
In attempting to adopt this work in our SSAI plugin, I realized that the best thing we were doing was to fire `nopreroll` and `nopostroll` when we load a new source. Doing this means contrib-ads treats everything as a midroll, which, from a UX perspective, is irrelevant.

So, I wonder if that behavior out to just be implicit in the `stitchedAds: true` scenario. It might make things simpler.

incompl:
@misteroneill What changes would that involve?

misteroneill:
I guess handling stitched ads as if `nopreroll` and `nopostroll` had always been fired.

misteroneill:
I have looked at `redispatch.handleEnded` and it seems the best path forward was for the `StitchedAdRoll` state to handle it and re-trigger `ended`. The fewer stitched-ads conditionals, the better.

Changing content during SSAI ads should work. I'm not sure contrib-ads can/should handle that, but in our plugin we should mostly handle it via middleware.

incompl:
I have been investigating changing sources during ads in contrib-ads and basically the issue is that there needs to be a way to differentiate "contentupdate due to ad loading" and "contentupdate due to content source change during ads". I'm not sure on what path to to take with that but input welcome. In the meantime, it's certainly not a blocker.

misteroneill:
Yeah, I took a quick look at our SSAI plugin and we do plenty of setup/teardown to handle changing in and out of SSAI sources, but it doesn't seem that we handle the case of changing during ad playback. Something to consider for the future.

test/karma.conf.js
```diff
@@ -38,15 +38,15 @@ module.exports = function(config) {
...
       'node_modules/sinon/pkg/sinon.js',
       'node_modules/video.js/dist/video.js',
-      'dist/videojs-contrib-ads.js',
-      'dist/videojs-contrib-ads.css',
-      'test/integration/lib/shared-module-hooks.js',
-      'test/dist/bundle.js',
+      {pattern: 'dist/videojs-contrib-ads.js', nocache: true},
+      {pattern: 'dist/videojs-contrib-ads.css', nocache: true},
+      {pattern: 'test/integration/lib/shared-module-hooks.js', nocache: true},
+      {pattern: 'test/dist/bundle.js', nocache: true},
```
misteroneill: This allows you to refresh the Karma Chrome instance and get the latest built files.

src/states/StitchedContentPlayback.js
```diff
@@ -0,0 +1,44 @@
...
+   * happen here, not in a constructor.
+   */
+  init() {
+
+    // Don't block calls to play in stitched ad players, ever.
+    this.player.ads._shouldBlockPlay = false;
+
+    // Tell redispatch not to intercept ended events. This should prevent
+    // the firing of readyforpostroll.
+    this.player.ads._contentHasEnded = true;
```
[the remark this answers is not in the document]
misteroneill: Sounds good to me.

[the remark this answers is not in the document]
misteroneill: Seems like we can add a check in redispatch for when we see `ended` with stitched ads, but we do still want to get an `adended` event so the states can handle that and get us out of ad mode before re-triggering `ended`. Updating soon.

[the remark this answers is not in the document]
incompl: If we return early in redispatch, it will allow the original ended event to not be cancelled or redispatched and you won't need to retrigger it at all.

examples/stitched-ad-plugin/plugin.js
```diff
@@ -0,0 +1,68 @@
...
+videojs.registerPlugin('exampleStitchedAds', function(options) {
+  var player = this;
+
+  options = options || {};
+  options.stitchedAds = true;
+
+  // Initialize contrib-ads.
+  player.ads(options);
+
+  player.on('canplay', function() {
```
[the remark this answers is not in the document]
misteroneill: It's mainly so that you can replay the example video.

src/redispatch.js
```diff
@@ -86,7 +86,7 @@ const handleEnded = (player, event) => {
     }
 
   // Prefix ended due to content ending before postroll check
-  } else if (!player.ads._contentHasEnded) {
+  } else if (!player.ads._contentHasEnded && !player.ads.stitchedAds()) {
```
incompl: Do you need it to go into the `if (player.ads.isInAdMode()) {` block above when using stitched ads? If not, could do a

```js
if (player.ads.stitchedAds()) {
  return
}
```

below `const handleEnded = (player, event) => {`

misteroneill: Yeah, it does need to go into that block because `StitchedAdRoll` needs to see `adended` to call `endLinearAdMode` and then re-trigger `ended`. Otherwise, you can end in an ad state, which is probably not desirable.

incompl: It seems strange to cancel ended so that you can retrigger it. What if you did the above, and then handle ended event by leaving ad mode?

misteroneill: I thought it might be weird to see an `ended` event during ad mode. If I were writing an integration, I would expect that `ended` did not happen during ad mode.

incompl: Ah so you want the sequence to be adended -> leave ad mode -> ended rather than ended -> leave ad mode. That makes sense.

incompl:
LGTM

PR #417, opened 2018-08-07
check shouldPlayContentBehindAd-for live only

rtezera1:
- Make sure `shouldPlayContentBehindAd` only returns true for `liveCuePoints`


test/integration/test.ads.js
```diff
@@ -774,41 +774,66 @@ QUnit.test('shouldPlayContentBehindAd', function(assert) {
   this.player.duration = function() {return Infinity;};
   videojs.browser.IS_IOS = true;
   videojs.browser.IS_ANDROID = true;
+  this.player.ads.settings.liveCuePoints = true;
```
incompl: true is the default, is it needed to set this?

rtezera1: good catch!

PR #420, opened 2018-08-13
check shouldPlayContentBehindAd for live only

rtezera1:
Make sure `shouldPlayContentBehindAd` only returns `true` for liveCuePoints


incompl:
LGTM

PR #421, opened 2018-08-20
fix: Avoid multiple-registration warning messages by accepting only the first contrib-ads per context.

misteroneill:


misteroneill:
This should also go into the same release as #415.

src/register.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @return {boolean}
+ *         If `true`, has the plugin. `false` otherwise.
+ */
+const hasAdsPlugin = () => {
+  if (videojs.getPlugin) {
+    return Boolean(videojs.getPlugin(NAME));
+  }
+
+  const Player = videojs.getComponent('Player');
```
incompl: I think a comment would be useful here, I'm not sure why this check is needed.

src/register.js
```diff
@@ -0,0 +1,73 @@
+import videojs from 'video.js';
+import playMiddlewareFeature from './playMiddleware.js';
+
+const {playMiddleware, isMiddlewareMediatorSupported} = playMiddlewareFeature;
+
+// The plugin name.
+const NAME = 'ads';
```
incompl: I recognize it's subjective, but extracting this to a variable doesn't seem valuable to me. The string is not likely to change and I don't see this as providing extra safety in this case. Importing it in other files seems tedious.

PR #422, opened 2018-08-22
chore: Remove unneeded onDispose handling

incompl:
This isn't necessary because on dispose the plugin is abandoned. It's state doesn't matter anymore because it won't be used. Any new players will have newly-initialized plugins with their initial state set correctly.


pagarwal123:
LGTM

pagarwal123:
QA-LGTM

PR #423, opened 2018-08-22
fix: Middleware log message

incompl:
Fixes a bug where the "registered middleware" log message would trigger after dispose/reinit even though middleware is not reregistered in this case.

Also fixes a bug where `initCancelContentPlay` would always be called if the debug setting was false. Not sure if this had functional consequences but it's clearly unintended.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -83,11 +83,7 @@ const contribAdsPlugin = function(options) {
   // will disallow calling play once play blocking is lifted)
   // The middleware must also be registered outside of the plugin,
   // to avoid a middleware factory being created for each player
-  if (isMiddlewareMediatorSupported() && settings.debug) {
-    // We log the debug message here as the plugin settings are available here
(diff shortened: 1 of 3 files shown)


src/plugin.js
```diff
@@ -80,11 +80,7 @@ const contribAdsPlugin = function(options) {
   // will disallow calling play once play blocking is lifted)
   // The middleware must also be registered outside of the plugin,
   // to avoid a middleware factory being created for each player
-  if (isMiddlewareMediatorSupported() && settings.debug) {
-    // We log the debug message here as the plugin settings are available here
-    videojs.log('ADS:', 'Play middleware has been registered with videojs');
-  } else {
-    // Register the cancelContentPlay feature on the player
+  if (!isMiddlewareMediatorSupported()) {
```
ldayananda: 😓

sirijami:
LGTM:QA

PR #424, opened 2018-08-23
fix: Remove hack in beforepreroll contentchanged handling

incompl:
I thought this was needed, but it turned out to workaround a bug in my integration. Cleaning up the upstream.


rtezera1:
QA:PASS

PR #433, opened 2018-09-06
chore: lint to code to vjs 7 standards

BrandonOCasey:
This removes the eslint configuration in favor of using `videojs-standard` which all of our other projects use. This pr is waiting for: https://github.com/videojs/standard/pull/36 to be merged so that we can turn the jsdoc rules off, even though they are just warnings and won't cause the build to fail.


misteroneill:
It looks like this needs an update to `.travis.yml` to no longer run tests in Node 4 because videojs-standard does not work there.

PR #449, opened 2018-11-14
Remove vjs-ad-loading class from player if vast response is empty

mimse:
If the vast request was empty the vjs-ad-loading class is not removed causing the video to be hidden behind an overlay

### DIFF
--- a/src/states/Preroll.js
+++ b/src/states/Preroll.js
@@ -228,6 +228,7 @@ export default class Preroll extends AdState {
     // Resume to content and unblock play as there is no preroll ad
     this.contentResuming = true;
     player.ads._shouldBlockPlay = false;
+    player.removeClass('vjs-ad-loading');
 
     // Play the content if we had requested play or we paused on 'contentupdate'
     // and we haven't played yet. This happens if there was no preroll or if it


incompl:
Do you have an example I could look at to help identify the issue?

mimse:
Hi, @incompl 

I enabled debug mode and this is the state changes that are happening.

image

Will try to make an example for you to test with.

mimse:
Hi again @incompl 

As you can see on this screenshot the vjs-ad-loading class is never removed.

image

To reproduce the issue download the src code from this link

https://drive.google.com/file/d/11EWqbimNlzLCEpJj4LtBZQhzXHJgLKTs/view?usp=sharing

Clicking play and reload the page until you get an empty VAST response.

incompl:
Ah, it does appear that the plugin is not aware that it transitioned beyond prerolls and that is why the preroll cleanup never happened. In searching your example for the root cause, I found that the cause is autoplay. The autoplay value "true" does not work with this plugin (basically because vjs middleware can't handle play events from the video element). There are two alternatives to address this: 1) call play manually instead of using the autoplay attribute or 2) use the new [string values for the autoplay option](https://docs.videojs.com/tutorial-options.html#autoplay).

mimse:
I´ll try that tomorrow. Thank you for looking into this.

mimse:
After changing how the autoplay option was set. I could not reproduce the issue. 

Thank you for helping with this, I´ll close the PR.

incompl:
Glad I could help!

Issue #450, opened 2018-11-23
Ad not playing when changing source while an ad is playing


mimse:
### Description

Changing source during an ad will cause the plugin state to be faulty.

### Steps to Reproduce

Use example basic-ad-plugin.

* Click play
* While ad is playing set a new source

#### Expected Results

The ad should play after setting a new source while an ad is playing

#### Actual Results

The content starts to play instead of an ad when setting a new source while an ad is playing

### Reproducing Unit Test

Please include a unit test that fails in the most recent version of videojs-contrib-ads but should pass according to your expected results. This will help us understand what the root cause is and separate ad plugin bugs from issues with this project. A unit test can be provided like below:

```js
import QUnit from 'qunit';
import {Midroll} from '../../../src/states.js';
import adBreak from '../../../src/adBreak.js';

QUnit.test('starts an ad break on init', function(assert) {
  this.midroll = new Midroll(this.player);
    
  this.midroll.init(this.player);
  assert.equal(this.player.ads.adType, 'midroll', 'ad type is midroll');
  assert.equal(this.adBreakStartStub.callCount, 1, 'ad break started');
});
```

This example was pulled from [test/states/test.Midroll.js](../test/states/test.Midroll.js).

### Versions

**videojs-contrib-ads version**: 6.6.1

**Video-js version: 7.3.0

### Platforms

**Browsers**: Chrome v70

**OS/Devices**: macOS Mojave

mysuf:
I already reported this in #445, but its better to have it in separate ticket..

PR #454, opened 2018-12-06
Fixing issue changing source while playing a preroll

mimse:
This PR fixes this videojs/videojs-contrib-ads#450 issue I created.

### DIFF
--- a/examples/basic-ad-plugin/app.js
+++ b/examples/basic-ad-plugin/app.js
@@ -99,6 +99,13 @@
 
   document.querySelector('form').addEventListener('submit', function(event) {
     event.preventDefault();
-    player.src(document.querySelector('input[type="url"]').value);
+    if (player.ads.inAdBreak()) {
+      // if in an adbreak while change source, call endLinearAdMode
+      player.ads.endLinearAdMode();
+      player.src(document.querySelector('input[type="url"]').value);
+    } else {
+      player.src(document.querySelector('input[type="url"]').value);
+    }
+    
   });
 })();
--- a/src/states/Preroll.js
+++ b/src/states/Preroll.js
@@ -1,6 +1,6 @@
 import videojs from 'video.js';
 
-import {AdState} from '../states.js';
+import {AdState, BeforePreroll} from '../states.js';
 import adBreak from '../adBreak.js';
 
 /*
@@ -222,6 +222,13 @@ export default class Preroll extends AdState {
     } else {
       this.noPreroll();
     }
+  }
+
+  /**
+   * If content changes while in Preroll state, transition to BeforePreroll
+   */
+  onContentChanged() {
+    this.transitionTo(BeforePreroll);
   }
 
   resumeAfterNoPreroll(player) {


incompl:
Hi there! This is a really important feature so I spent some time testing it and exploring how things work.

I work on an ad integration for which source changes during ads work without this change, so I was really curious why it worked there but didn't work in the example integration. I found that, in short, source change during ads works as long as ad mode is ended when the source changes. This works because ad mode ending puts you in the ContentPlayback state, and then the `contentchanged` handling that already exists takes care of the rest.

According to the videojs-contrib-ads contract and documentation, the ad integration is responsible for calling endLinearAdMode. So I started looking for a way to address this in `example-plugin.js`. I thought it was probably possible with videojs middleware but didn't quite get it working.

Then I realized your PR does indeed call endLinearAdMode in app.js, and that should be enough. So I copied your PR with just the change to app.js and it seems to work.

So, in short: it looks like your change to `app.js` is sufficient and the change to `Preroll.js` is not necessary.

I'd be willing to accept that change, but since endLinearAdMode should be invoked by the ad plugin, a solution that is constrained within `example-plugin.js` may be a more useful example to anyone who wants to use this project and get the benefit of source changing during ads.

mimse:
Hi @incompl,

Thanks for having a look at my PR.

I have been testing a lot the last week and haven't been able to get it working as defined in the documentation.

This is the events I´m seeing running the example with my modifications, 
image

And here without my changes to Preroll state
image

As you can see, it plays an ad after the source is changed. Which is correct. But it plays as content and not as an Ad.

Also getting VIDEOJS: WARN: Unexpected startLinearAdMode invocation (Preroll) in console.

incompl:
I do think the trick to this is going to be calling endLinearAdMode at the right time on source change, not adding a state transition. If you're getting "Unexpected startLinearAdMode" it's a sign that the change to the state transition logic is creating an unexpected flow. I don't believe that adding a state transition is going to be likely to address this.

incompl:
I created a PR to try to address it by ending linear ad mode in middleware. Could you try it out? https://github.com/videojs/videojs-contrib-ads/pull/455

mimse:
Hi @mysuf 

That is the thing. I do not have the full overview of this codebase to know if I'm breaking stuff in other places or this needs to be added to other types of adbreaks to.

This is why I created this PR :-)

mysuf:
Hi. I only pointed that out. If transition properly resets current state before leaving (e.g. clearing timeouts, pending events etc), then I'm voting for transition. If not, then @incompl 's middleware PR is probably better way.

incompl:
I think this PR has potential to be a better solution, but it's not there at this point.

Requiring `endLinearAdMode` to be called by the external code that is changing the source (in this example, `app.js` is not tenable in practice: we don't have control over the code that changes sources and we can't ask that every 3rd party piece of code that changes sources contain videojs-contrib-ads support. Speaking as the developer of production ad plugins that uses videojs-contrib-ads, this would not be a criteria I'd be able to accept and deploy to our users. This was the main goal I set out to resolve in my PR. Maybe there is a better way than using middleware, like calling endLinearAdMode in the state transition code, but I haven't been able to find one yet.

A different issue is that when I test this code, there is a delay of a couple seconds before the new ad/content plays. I'm not sure why, but this suggests to me that something isn't cleaned up and it doesn't work until a timeout fires. In order to have confidence that the code functions as intended, I'd need to see that delay explained or fixed.

I wish I had time to dig into this more, but it's a doozy of an issue. For now, that's where I'm at.

mysuf:
I think that  (I dont know this plugin like you do @incompl) the best solution would be contrib-ads event handler on source middleware since plugin registration. Then you never miss any event and you'll get more options. Before contrib-ads getAds init, it could be basic event handler watching for load events. With getAds, it could be replaced with full event handler. Another step would be reset() method that will properly reset contrib-ads to default state (like when you initialize plugin first time) called when contrib-ads will be in other than default state.

mimse:
> Requiring endLinearAdMode to be called by the external code that is changing the source (in this example, app.js is not tenable in practice: we don't have control over the code that changes sources and we can't ask that every 3rd party piece of code that changes sources contain videojs-contrib-ads support.

This is what videojs-ima is doing.

https://github.com/googleads/videojs-ima/blob/e64564b5fe4b5ba9497c1226c90cbd71eb2a2fb7/src/player-wrapper.js#L614

Not saying it is the right way of doing it. But that is where I got the idea from :-)

zgzong:
The main issue is whether the videojs-contrib-ads should support player zapping during an ad? If it is on the roadmap, I agree @mysuf the videojs-contrib-ads need be updagted. Otherwise #455 workaround is good to make sure contrib-ads in correct state.

PR #455, opened 2018-12-07
[WIP] Experiment to get source change during ads working in example integration

incompl:


mimse:
Hi @incompl 

Just tested your solution. This is what I get in Chrome 71 on mac

image

I click source, it stops the ad, but it does not continue to a new ad, instead, it stops.

incompl:
I believe all it needs is a call to `player.play()` depending on the desired behavior

mimse:
It works when I added play. Thank you for the solution. I would recommend adding this to your documentation or as an example.
Don´t think anyone would be able to figure this out on them selfs :-)

If my solution doesn't have any other side effects, I still think it might be a better solution. Because as a user of the plugin you would not need to add anything to make it work.

PR #461, opened 2019-01-14
refactor: Add internal registration system for states to work around circular dependency issues.

BrandonOCasey:
Currently there is a circular dependency issue that locks this repository to an old version of rollup, and thus an old version of the generator. This pull request seeks to remedy that. I implemented a registration system similar to what we have for Components in Video.js, but it will be for States here.

### DIFF
--- a/src/states.js
+++ b/src/states.js
@@ -1,29 +1,25 @@
-/*
- * This file is necessary to avoid this rollup issue:
- * https://github.com/rollup/rollup/issues/1089
- */
-import State from './states/abstract/State.js';
-import AdState from './states/abstract/AdState.js';
-import ContentState from './states/abstract/ContentState.js';
-import Preroll from './states/Preroll.js';
-import Midroll from './states/Midroll.js';
-import Postroll from './states/Postroll.js';
-import BeforePreroll from './states/BeforePreroll.js';
-import ContentPlayback from './states/ContentPlayback.js';
-import AdsDone from './states/AdsDone.js';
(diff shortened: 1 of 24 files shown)


src/states.js
```diff
@@ -1,29 +1,25 @@
-/*
- * This file is necessary to avoid this rollup issue:
- * https://github.com/rollup/rollup/issues/1089
```
misteroneill: Was this a different circular dependency than this issue?

BrandonOCasey: Yeah, this issues was only in rollup 0.50.1, after that another issue pops up. That was the one that was fixed in this pull request. rollup would still build the files correctly but would append `$$1` to most of the states to avoid the circular dependency problem. Since most of the tests use the constructor name this caused them to fail.

src/states/BeforePreroll.js
```diff
@@ -42,6 +44,8 @@ export default class BeforePreroll extends ContentState {
    * content playback is blocked by the ad plugin.
    */
   onPlay(player) {
+    const Preroll = States.getState('Preroll');
```
incompl: If this is what rollup needs I guess you gotta do it. My only input is that putting the getState const at the top of the file makes the most sense for keeping the code tidy.

BrandonOCasey: Yeah I tried a few ways to keep it like that, but then it became impossible to `getState` them all in the correct order. So one of the `getState` calls would return `undefined` and break everything. I think we have to do it at runtime.

PR #462, opened 2019-01-14
chore: Update development tooling

BrandonOCasey:
Based on #461


misteroneill:
@BrandonOCasey The tests are failing in Chrome 71 for me locally due to `play()` being called without user interaction - they pass on `master`. Maybe some kind of Karma difference?

misteroneill:
Actually, it seems that it was user error. I am no longer seeing the test failures.

PR #470, opened 2019-03-04
fix: Clean up player when a preroll fails and prevent uncaught play promise exceptions.

misteroneill:
When using autoplay and receiving an `adserror` event, the `Preroll` state will cause an uncaught play promise exception and will never remove the `vjs-ads-loading` class, which can result a player that looks broken.

### DIFF
--- a/src/states/Preroll.js
+++ b/src/states/Preroll.js
@@ -154,7 +154,7 @@ class Preroll extends AdState {
     const player = this.player;
 
     if (this.adsReady && !player.ads.inAdBreak() && !this.isContentResuming()) {
-      player.clearTimeout(this._timeout);
+      this.clearTimeout(player);
       player.ads.adType = 'preroll';
       this.waitingForAdBreak = false;
       adBreak.start(player);
@@ -227,16 +227,21 @@ class Preroll extends AdState {
   }
 
   resumeAfterNoPreroll(player) {
+
     // Resume to content and unblock play as there is no preroll ad
     this.contentResuming = true;
     player.ads._shouldBlockPlay = false;
+    this.cleanupPartial(player);
 
     // Play the content if we had requested play or we paused on 'contentupdate'
     // and we haven't played yet. This happens if there was no preroll or if it
     // errored, timed out, etc. Otherwise snapshot restore would play.
-    if (player.paused() &&
-        (player.ads._playRequested || player.ads._pausedOnContentupdate)) {
-      player.play();
+    if (player.paused() && (player.ads._playRequested || player.ads._pausedOnContentupdate)) {
+      const playPromise = player.play();
+
+      if (playPromise) {
+        playPromise.then(() => {}).catch(() => {});
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/states/Preroll.js
```diff
@@ -247,12 +252,26 @@ class Preroll extends AdState {
...
-    player.clearTimeout(this._timeout);
+    this.clearTimeout(player);
   }
 
+  /*
+   * Clear the preroll timeout and nulls out the pointer.
+   */
+  clearTimeout(player) {
+    player.clearTimeout(this._timeout);
+    this._timeout = null;
```
misteroneill: This mainly makes testing a little easier.

test/unit/states/test.Preroll.js
```diff
@@ -22,8 +23,9 @@ QUnit.module('Preroll', {
       },
       setTimeout: () => {},
       clearTimeout: () => {},
-      addClass: () => {},
-      removeClass: () => {},
+      addClass: (name) => this.classes.push(name),
+      removeClass: (name) => this.classes.splice(this.classes.indexOf(name), 1),
+      hasClass: (name) => this.classes.indexOf(name) !== -1,
```
misteroneill: Quick and dirty mock for class manipulation methods.

src/states/Preroll.js
```diff
@@ -227,16 +227,21 @@ class Preroll extends AdState {
   }
 
   resumeAfterNoPreroll(player) {
+
     // Resume to content and unblock play as there is no preroll ad
     this.contentResuming = true;
     player.ads._shouldBlockPlay = false;
+    this.cleanupPartial(player);
```
misteroneill: This seems to address the player being placed in a bad state (e.g. loading spinner) in some situations.

src/states/Preroll.js
BrandonOCasey: Is this comment referring to setting `_timeout` to `null`? or this function as a whole? Would be good to add a comment to the function regarding it's usage for testing (if we want to keep it) as I could see this being refactored out easily.

src/states/Preroll.js
```diff
@@ -227,16 +227,21 @@ class Preroll extends AdState {
...
     // and we haven't played yet. This happens if there was no preroll or if it
     // errored, timed out, etc. Otherwise snapshot restore would play.
-    if (player.paused() &&
-        (player.ads._playRequested || player.ads._pausedOnContentupdate)) {
-      player.play();
+    if (player.paused() && (player.ads._playRequested || player.ads._pausedOnContentupdate)) {
+      const playPromise = player.play();
+
+      if (playPromise) {
+        playPromise.then(() => {}).catch(() => {});
```
BrandonOCasey: We should probably copy what we do in video.js here. Not sure if this is actually an issue or not though. https://github.com/videojs/video.js/blob/master/src/js/utils/promise.js#L25

misteroneill: Reasonable!

PR #473, opened 2019-03-11
fix: Remove duplicate playing listener from redispatch.

BrandonOCasey:
`playing` is an event that already exists in `VIDEO_EVENTS`, this may have caused us to fire two `playing` or `adplaying` events sometimes.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -82,7 +82,7 @@ const contribAdsPlugin = function(options) {
   const videoEvents = [];
 
   // dedupe event names
-  VIDEO_EVENTS.concat(['firstplay', 'loadedalldata', 'playing']).forEach(function(eventName) {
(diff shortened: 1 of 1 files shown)


src/plugin.js
```diff
@@ -79,11 +79,14 @@ const contribAdsPlugin = function(options) {
...
   // avoiding the complexity for common usage
-  const videoEvents = VIDEO_EVENTS.concat([
-    'firstplay',
-    'loadedalldata',
-    'playing'
-  ]);
+  const videoEvents = [];
+
+  // dedupe event names
+  VIDEO_EVENTS.concat(['firstplay', 'loadedalldata', 'playing']).forEach(function(eventName) {
```
misteroneill: Why not remove `playing` here entirely? We know it's a dupe.

PR #476, opened 2019-04-10
feat: add pageUrl macro

mister-ben:
Adds a `player.pageUrl` macro. This is intended as the URL at which the player has been embedded, whether directly or via an iframe. The name of the macro hopefully makes it clear that this is not an embed URL for the player itself.

If the player is in an iframe, this resolves to `document.referrer`, otherwise to `window.location.href`.

A test is only included for the iframe case for now, as the tests themselves run in an iframe.

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -65,6 +65,8 @@ export default function adMacroReplacement(string, uriEncode, customMacros) {
   macros['{mediainfo.name}'] = this.mediainfo ? this.mediainfo.name : '';
   macros['{mediainfo.duration}'] = this.mediainfo ? this.mediainfo.duration : '';
   macros['{player.duration}'] = this.duration();
+  // When not in an iframe, window.location and window.parent.location are the same
+  // Within an iframe, window.parent.location is different or inaccessible
   macros['{player.pageUrl}'] = (window.location !== window.parent.location) ?
     document.referrer :
     window.location.href;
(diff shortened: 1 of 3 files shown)


mister-ben:
Failing tests are unrelated

misteroneill:
Failing tests are related to Firefox 66. I believe @alex-barstow is fixing that in another PR>

src/macros.js
```diff
@@ -65,6 +65,9 @@ export default function adMacroReplacement(string, uriEncode, customMacros) {
   macros['{mediainfo.name}'] = this.mediainfo ? this.mediainfo.name : '';
   macros['{mediainfo.duration}'] = this.mediainfo ? this.mediainfo.duration : '';
   macros['{player.duration}'] = this.duration();
+  macros['{player.pageUrl}'] = (window.location !== window.parent.location) ?
```
misteroneill: Maybe a comment here explaining that this conditional is used to detect iframes? It may be unfamiliar to some.

gkatsev: I think that secure iframes throw if you access parent location, we'd want to double check and wrap in a try/catch if so

gkatsev: Videojs has a isInIframe helper on the Dom until object

bcbclifford: on chrome/ff/safari, `window.parent.location` doesn't throw, although `window.parent.location.href` does.

mister-ben: didn't realise we had isInFrame. will use that instead then.

PR #486, opened 2019-05-10
Fix: on auto-play don't request playback for stitched ads as well as when playback already started

bcdarius:
Some techs, e.g. Flash may re-trigger `canplay` after playback has begun which can result in requesting playback when playback is already progressing. When assuming auto-play, playback should be requested only if it hasn't started. We also don't need to handle auto-playback for stitched ads because linear ads in such cases are stitched into the content.


bcdarius:
@misteroneill That's a good point, for stitched ads this handler is redundant.

bcdarius:
@misteroneill Is this good to go? Thanks

PR #513, opened 2020-04-15
Fix: Player does not trigger 'play' and 'pause' events first time after AdBlocker block Ads

marcodeltorob:
**Problem Description**

When Preroll is blocked due to AdBlocker the plugin will trigger an `adtimeout` event then content will play without any event trigger like `play` or `contentplay` or transition to `ContentPlayback` state. So, the first event triggered if we call `player.play()` or `player.pause()` will be prefixed with `content` (`contentpause` or `contentplay`) and plugin will transition from `Preroll -> ContentPlayback` because we are in `Preroll` stuck. 

**Solution Proposal**

When we have and `adtimeout` in `Preroll` state transition to `ContentPlayback`

### DIFF
--- a/src/states/Preroll.js
+++ b/src/states/Preroll.js
@@ -210,8 +210,11 @@ class Preroll extends AdState {
    */
   onAdTimeout(player) {
     this.afterLoadStart(() => {
+      const ContentPlayback = States.getState('ContentPlayback');
+
       player.ads.debug('adtimeout (Preroll)');
       this.resumeAfterNoPreroll(player);
+      this.transitionTo(ContentPlayback);
     });
   }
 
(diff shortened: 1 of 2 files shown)


marcodeltorob:
What `resumeAfterNoPreroll` does is to set `contentResuming` to true and unlock playback with `player.ads._shouldBlockPlay = true` also remove the `timeout` function and the CSS (`vjs-ad-loading` & `vjs-ad-content-resuming`).

In a normal scenario, the plugin will transition from `Preroll -> ContentPlayback` after the Ad brake but here the plugin is getting stuck in `Preroll` state so that is why inside the `onAdTimeout()` I'm proposing a transition to  `ContentPlayback` in case of the `adtimeout` is triggered since due to the AdBlocker the plugin does not fire more events and is not transitioning until the first manual `player.play()` and that is why the player is sending a `contentplay` instead of on `play` event.

gkatsev:
I'm a bit worried that this will change the behavior of adtimeout in the case adblockers aren't available. Like if the ad response is slow or the response contains no ads.

misteroneill:
As discussed offline, we think this carries some risk of creating other undesirable behaviors and will close for now. If this issue comes back up, we can revisit!

PR #518, opened 2020-07-10
Make video.js as peerDependencies

ckybonist:
Please see https://github.com/googleads/videojs-ima/pull/912

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -890,6 +890,93 @@
         "shx": "^0.3.2"
       }
     },
+    "@videojs/http-streaming": {
+      "version": "1.13.2",
+      "resolved": "https://registry.npmjs.org/@videojs/http-streaming/-/http-streaming-1.13.2.tgz",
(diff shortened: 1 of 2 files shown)


gkatsev:
This has been something that we've been thinking about for a while. Generally, the issue of having multiple versions in the bundle goes away if you use compatible versions of Video.js, which also why our version range is fairly unrestrictive.
Historically, we've avoided peerDeps because the way npm worked made it not usable for us. I believe a lot of those issues have been addressed on the npm side but we haven't had the chance to take a look at the implications of doing this.

ckybonist:
Okay. Just workaround this issue by setting webpack alias path like:

```js
alias: {
    src: path.resolve(__dirname, '../../src'),
    assets: path.resolve(__dirname, '../../assets'),
    'video.js': path.resolve(__dirname, '../../node_modules/video.js')
  }
```

PR #525, opened 2021-01-20
fix: Ad plugin stuck in Preroll when adserror due empty VAST

marcodeltorob:
**Problem Description**

When the Ads plugin throws an `adserror (Preroll)` due `AdError 1009: The VAST response document is empty` is getting stuck in the `Preroll` state and not switching to `ContentPlayback`  because of this behavior the Ads plugin is dispatching `timeupdate` as `contenttimeupdate`. So, in and `adserror` we want to transition to `ContentPlayback` with a `contentresumed` if the player is actually playing content.

### DIFF
--- a/src/states/Preroll.js
+++ b/src/states/Preroll.js
@@ -240,9 +240,16 @@ class Preroll extends AdState {
       const playPromise = player.play();
 
       if (playPromise && playPromise.then) {
-        playPromise.then(null, (e) => {});
+        playPromise.then(null, (e) => { });
       }
+      // When Ad plugin throws and adserror (Preroll) after initialation due AdError 1009: The VAST response document is empty
+      // ad plugin state won't transition from Preroll -> ContentPlayback and will be stuck in AdMode
+      // so any event like timeupdate will be redispatched with the prefix content (contenttimeupdate)
+    } else if (!player.paused() && !player.ads.inAdBreak() && !player.ads.isAdPlaying() && player.ads.isContentResuming()) {
+      // trigger contentresumed to switch from Preroll -> ContentPlayback
+      player.trigger('contentresumed');
     }
+
   }
 
   /*
--- a/src/states/Preroll.js
+++ b/src/states/Preroll.js
@@ -143,10 +143,16 @@ class Preroll extends AdState {
     } else {
       this.afterLoadStart(() => {
         this.resumeAfterNoPreroll(player);
+        // When Ad plugin throws and adserror (Preroll) after initialation due AdError 1009: The VAST response document is empty
+        // ad plugin state won't transition from Preroll -> ContentPlayback and will be stuck in AdMode
+        // so any event like timeupdate will be redispatched with the prefix content (contenttimeupdate)
+        if (!player.paused() && !player.ads.inAdBreak() && player.ads.isContentResuming()) {
+          // trigger contentresumed to switch from Preroll -> ContentPlayback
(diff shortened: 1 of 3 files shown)


src/states/Preroll.js
```diff
@@ -240,9 +240,16 @@ class Preroll extends AdState {
...
       if (playPromise && playPromise.then) {
-        playPromise.then(null, (e) => {});
+        playPromise.then(null, (e) => { });
       }
+      // When Ad plugin throws and adserror (Preroll) after initialation due AdError 1009: The VAST response document is empty
+      // ad plugin state won't transition from Preroll -> ContentPlayback and will be stuck in AdMode
+      // so any event like timeupdate will be redispatched with the prefix content (contenttimeupdate)
+    } else if (!player.paused() && !player.ads.inAdBreak() && !player.ads.isAdPlaying() && player.ads.isContentResuming()) {
+      // trigger contentresumed to switch from Preroll -> ContentPlayback
+      player.trigger('contentresumed');
```
misteroneill, line 250: Is this something that could be done in an `onAdsError` callback?

marguinbc:
I don not believe this error should cause the the ad to fail. It is common for VMAP wrapper ads to return 1009, empty vast when one or more wrappers returns without a matching ad. This triggers the VMAP to look at the next wrapper, and so on. Killing the ad when the first 1009 occurs would be problematic as the ad would stop and not proceed to the next wrapper candidate.

marguinbc:
I'm suspecting an issue in the IMA SDK or Chrome as there are 3-4 threads about weird behavior with 1009 on the IMA SDK Google Group, like this one for example: https://groups.google.com/g/ima-sdk/c/nqWj9IZBJzk/m/76cy8255AAAJ

misteroneill:
Discussed offline. Since this looks like it's related to the IMA SDK, we probably don't want to work around it in the contrib-ads plugin (it should be agnostic as pertains to the ad provider).

misteroneill:
After doing some investigation on this myself, it looks like what happens sometimes is that the `playing` event occurs while the player is still in ad mode. When we go to resume from the `Preroll` state, the code that triggers playback doesn't do anything because the player is already playing.

I think we should detect this condition and, instead, trigger the `playing` event so that it doesn't get swallowed and content mode resumes properly.

misteroneill (changes requested):
I think the change is fine - it addresses the issue we've been seeing - but it looks like there's a test failing now. Also, can we add a test that reproduces the original issue?

test/integration/test.cancelContentPlay.js
```diff
@@ -110,8 +110,8 @@ QUnit.test("cancelContentPlay doesn\'t block play in content playback", function
   this.player.trigger('play');
   assert.strictEqual(pauseSpy.callCount, 1, 'pause should have been called');
   assert.strictEqual(
-    this.player.ads._cancelledPlay, true,
-    'cancelContentPlay is called while resuming'
```
misteroneill, line 114: So, simply changing this test suggests that the behavior of the library changed. Instead of changing the test, we should investigate why this behavior changed.

misteroneill (changes requested):

test/integration/test.cancelContentPlay.js
misteroneill: Following our Slack discussion, we may want to try firing both `play` and `playing` 🤔

misteroneill (approved):

PR #527, opened 2021-03-29
chore: Do not run tests on version

misteroneill:
Nothing should be merged unless tests are passing anyway.


brandonocasey (approved):
We don't do it in the generator, no reason to do it here.

PR #532, opened 2021-06-17
feat: Add `allowVjsAutoplay` option to add support for custom video.js autoplay settings

alex-barstow:
## Description
Video.js recently added new functionality for autoplay that triggers `autoplay-success` or `autoplay-failure` events depending on the result of the autoplay attempt. However, that functionality relies on the initial play promise being resolved/rejected naturally, so it is not compatible with the contrib-ads `callPlay()` middleware which terminates the first play request.

This PR adds an option to bypass that play termination if the player intends to autoplay. It does this by conditionally setting `_shouldBlockPlay` to `false` in the `BeforePreroll` state if the option is `true` and `player.autoplay()` is truthy.

### DIFF
--- a/docs/integrator/options.md
+++ b/docs/integrator/options.md
@@ -59,7 +59,7 @@ Use this to override detection of if the content video is a live stream. Live de
 ### allowVjsAutoplay
 
 Type: `boolean`
-Default Value: `videojs.options.normalizeAutoplay` || false
+Default Value: `videojs.options.normalizeAutoplay || false`
 
 Set this to `true` if you intend to use video.js's custom autoplay settings ("play", "muted", or "any"). It defaults to `true` if the videojs `normalizeAutoplay` option is `true` since `normalizeAutoplay` signals an intent to use `autoplay: "play"` behavior.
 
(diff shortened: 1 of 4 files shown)


misteroneill (approved):

docs/integrator/options.md
```diff
@@ -56,6 +56,13 @@ No Default Value
 
 Use this to override detection of if the content video is a live stream. Live detection checks if the duration is `Infinity` but there are cases when this check is insufficient.
 
+### allowVjsAutoplay
+
+Type: `boolean`
+Default Value: `false`
+
+Set this to `true` if you intend to use video.js's custom autoplay settings ("play", "muted", or "any").
```
gkatsev, line 64: Suggestion: Should we mention`normalizeAutoplay` here?

gkatsev, line 64: Thought: should we assume `allowVjsAutoplay` if `normalizeAutoplay` is set?

gkatsev: Specifically, since they're basically intended to work together.

alex-barstow: Yup, I'll add a mention

alex-barstow: Yeah, that makes sense to me

alex-barstow: Updated

gkatsev (approved):

PR #538, opened 2022-09-23
fix: Don't break on nested macros with undefined parent

mister-ben:
Currently, a macro like `{pageVariable.foo.bar}` will cause an error if `foo` is not defined, as there's an attempt to access `foo.bar`.

Adds a check for whether each level is defined within the pageVariable string replacement.


wseymour15 (approved):
LGTM

Issue #488, opened 2019-05-17
Specific test failing with Video.js 7.5.x


misteroneill:
[This test](https://github.com/videojs/videojs-contrib-ads/blob/master/test/integration/test.snapshot.js#L293-L324) fails with Video.js 7.5.x

PR #541, opened 2023-01-30
chore: remove deprecated functionality

wseymour15:
**Problem Description**

Some of the functionality in this repo is deprecated in VJS v8, so we are updating contrib-ads to depend on v8 and update the deprecated functionality. 

Additionally, an integration test was failing if you attempted to depend on video.js version 7.5.0 or higher (related to [this issue](https://github.com/videojs/videojs-contrib-ads/issues/488)), so I updated that test to pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
   ],
   "dependencies": {
     "global": "^4.3.2",
-    "video.js": "^6 || ^7"
+    "video.js": "^7 || ^8"
   },
   "generator-videojs-plugin": {
     "version": "7.6.1"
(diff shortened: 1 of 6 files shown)


dzianis-dashkevich (approved):

package.json
```diff
@@ -77,7 +77,7 @@
   ],
   "dependencies": {
     "global": "^4.3.2",
-    "video.js": "^6 || ^7"
+    "video.js": "^7 || ^8"
```
mister-ben, line 80: The other changes would only be compatible with 8

wseymour15: Good call, I removed the v7 option here

mister-ben:
As this will need a major, it's also be a good opportunity to move video.js from dependencies to devDependencies and peerDependencies.

PR #544, opened 2023-03-29
feat: refactor macro logic and add support for disableDefaultMacros and macroNameOverrides parameters

alex-barstow:
## Description
This PR makes the following changes:
- Refactor `macros.js` by breaking logic groups into separate functions
- Adds two new macro features:
  - **disableDefaultMacros** - Allows integrators to use `adMacroReplacement()` for _only_ customMacros, skipping replacement of the default macros (ex. `{player.id}`)
  - **macroNameOverrides** - Allows integrators to customize the names of any default macros, including dynamic macros. For example, they can replace `{player.id}` with `{{{PLAYER_ID}}}`, `{tcf.gdprApplies}` with `{{GDPR}}`, etc
- Update docs and tests

The new options are passed as properties of the existing `customMacros` object:

```js
const customMacros = {
  disableDefaultMacros: false,
  macroNameOverrides: {
    '{player.id}': '{{PLAYER_ID}}',
  },
  '{{yourCustomMacro}}': 'yourValue',
  '{{otherCustomMacro}}': 'otherValue'
};

const adTagUrlWithMacrosReplaced = player.ads.adMacroReplacement(adTagUrl, true, customMacros);
```

## Notes
One possible objection to this approach is that it introduces "magic properties" to `customMacros` that behave differently than other properties. They are implemented in such a way that there should not be any adverse side-effects (they are deleted from the `customMacros` object before it is processed), and I'm personally fond of the simplicity of this approach. It doesn't add any new parameters to `adMacroReplacement()` or the need for additional method calls.

Ideally, we would do something like this, but unfortunately it isn't backward-compatible:

```js
const macroCustomizationOptions = {
  disableDefaultMacros: false,
  macroNameOverrides: {
    '{player.id}': '{{PLAYER_ID}}',
  },
  customMacros: {
    '{{yourCustomMacro}}': 'yourValue',
    '{{otherCustomMacro}}': 'otherValue'
  }
};

const adTagUrlWithMacrosReplaced = player.ads.adMacroReplacement(adTagUrl, true, macroCustomizationOptions);
```
 
If there are very strong objections to my approach, here are two other possible approaches:

**1. Allow the `customMacros` argument to be a function that returns an object:**

```js
player.ads.adMacroReplacement(adTagUrl, true, () => {
  return {
    disableDefaultMacros: false,
    macroNameOverrides: {
      '{player.id}': '{{PLAYER_ID}}',
    },
    customMacros: {
      '{{yourCustomMacro}}': 'yourValue',
      '{{otherCustomMacro}}': 'otherValue'
    }
  };
})
```
This lets us adopt the ideal data structure in a backward-compatible way, but it adds a bit more conceptual complexity. It's also still very much a workaround.

**2. Add 2 new methods that can be called before `adMacroReplacement()`:**

```js
player.ads.disableDefaultMacros(true);
player.ads.macroNameOverrides({
 '{player.id}': '{{PLAYER_ID}}'
});
player.ads.adMacroReplacement(adTagUrl, true, customMacros)`
```
This is fairly clean, but it does require two separate method calls.


docs/integrator/macros.md
```diff
@@ -85,10 +94,52 @@ Since `gdprApplies` is a boolean and many ad servers expect the value as an int,
...
 If the player is in an iframe, a proxy will be added if any parent frame is detected to gain consent with the postmessage API. The CMP must be loaded first.
 
-## Default values in macros
+### Default values in macros
 
 A default value can be provided within a macro in the format `{MACRO=DEFAULT}`, in which case this value will be used where not resolvable e.g. `http://example.com/ad/{pageVariable.adConf=1234}` becomes `http://example.com/ad/1234` if `window.adConf` is undefined. If a blank default is given, there will be a blank param, `http://example.com/ad?config={pageVariable.adConf=}&a=b` becomes `http://example.com/ad?config=&a=b` if `window.adConf` is unset.
 
 [tcf]: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md
 [tcdata]: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#tcdata
 [referrer-policy]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy
```
mister-ben, line 103: It won't matter when rendered, but these link definitions are probably better at the end of the doc.

src/macros.js
```diff
@@ -131,24 +140,85 @@ export default function adMacroReplacement(string, uriEncode, customMacros) {
...
+  return string;
+};
+
+/**
+ *
+ * @param {string} string
+ *                 Any string with macros to be replaced
+ * @param {boolean} uriEncode
+ *                  A Boolean value indicating whether the macros should be replaced with URI-encoded values
+ * @param {object} customMacros
```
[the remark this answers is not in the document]
alex-barstow: This repo doesn't actually use jsdoc, so I'm not really following any strict convention. I just find the jsdoc comment format familiar and clear so I chose to adopt it here.

src/macros.js
```diff
@@ -33,86 +29,99 @@ const customFields = function(mediainfo, macros, customFieldsName) {
...
-  ['description', 'tags', 'reference_id', 'ad_keys'].forEach((prop) => {
-    if (this.mediainfo && this.mediainfo[prop]) {
-      macros[`{mediainfo.${prop}}`] = this.mediainfo[prop];
-    } else if (defaults[`{mediainfo.${prop}}`]) {
-      macros[`{mediainfo.${prop}}`] = defaults[`{mediainfo.${prop}}`];
-    } else {
-      macros[`{mediainfo.${prop}}`] = '';
-    }
-  });
+const getStaticMacros = function() {
```
misteroneill, line 77: suggestion/non-blocking: I would suggest this function take a `player` argument instead of expecting it to be invoked via `call()`, but that's very minor for an internal function.

src/macros.js
```diff
@@ -131,24 +140,85 @@ export default function adMacroReplacement(string, uriEncode, customMacros) {
...
+  // Get macros with defaults e.g. {x=y}, store the values in `defaults` and replace with standard macros in the string
+  const {defaults, modifiedString} = getDefaultValues(string);
+
+  string = modifiedString;
+
+  // Get all macro values
+  Object.assign(macros, getStaticMacros.call(this));
+  Object.assign(macros, getMediaInfoMacros(this.mediainfo, defaults));
+  Object.assign(macros, getTcfMacros(tcData));
+  Object.assign(macros, getPageVariableMacros(string, defaults));
```
misteroneill, line 214: nitpick/non-blocking: These could be a single `Object.assign()` call as it can take any number of arguments.

test/integration/test.macros.js
```diff
@@ -327,3 +327,142 @@ QUnit.test('tcfMacros', function(assert) {
 
   window.__tcfapi = oldtcf;
 });
+
+QUnit.test('default macros should not be replaced when disableDefaultMacros is true', function(assert) {
+  const string = '{player.id}';
+  const customMacros = {
+    disableDefaultMacros: true
```
misteroneill, line 334: suggestion/non-blocking: I would suggest that we have a test that ensures that "magic" properties of `customMacros` do not get replaced themselves. Something like:
```js
const string = 'disableDefaultMacros';
const result = this.player.ads.adMacroReplacement(string, false, {
  disableDefaultMacros: false
});

assert.strictEqual(result, string, 'disableDefaultMacros was not treated as a macro');
```

misteroneill (approved):
Overall 👍 

Added a couple minor suggestions that should not block this PR.

src/macros.js
alex-barstow: Lol yup good point

alex-barstow:
I'm going to merge then follow-up with a PR to address Pat's comments.

PR #548, opened 2023-04-06
fix: playback rate should be 1 while ad is playing

philjhale:
**Problem**

Ads can be played at a speed greater than 1.  See [this demo](https://players.brightcove.net/6415614376001/k2bS2hi9ri_default/index.html?videoId=6322666257112). Set the speed to x3 and wait for the mid-roll ad to play. Expected behaviour, play at x1 speed. Actual behaviour, plays at x3 speed.

**Changes**

* Sets playback rate to 1 during an ad. 
* Set playback rate to the previous rate once ad has finished.
* If the playback rate menu is visible in the player, hide it during the ad and show it again when the ad finishes.

Here is a [demo of the fix](https://players.brightcove.net/6415614376001/hnLwSyWKd_default/index.html?videoId=6322666257112).

### DIFF
--- a/src/adBreak.js
+++ b/src/adBreak.js
@@ -42,8 +42,12 @@ function start(player) {
   // Ensure ads are watched at x1 speed and speed cannot be changed for the duration of the ad
   player.ads.preAdPlaybackRate_ = player.playbackRate();
   player.playbackRate(1);
-  if (player.controlBar.playbackRateMenuButton) {
+  if (player.controlBar.playbackRateMenuButton.playbackRateSupported &&
+      !player.controlBar.playbackRateMenuButton.hasClass('vjs-hidden')) {
     player.controlBar.playbackRateMenuButton.hide();
+    player.ads.showPlaybackMenuOnAdEnd_ = true;
+  } else {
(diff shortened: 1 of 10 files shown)


src/adBreak.js
```diff
@@ -38,6 +38,17 @@ function start(player) {
   // This removes the native poster so the ads don't show the content
   // poster if content element is reused for ad playback.
   player.ads.removeNativePoster();
+
+  // Ensure ads are watched at x1 speed and speed cannot be changed for the duration of the ad
+  player.ads.preAdPlaybackRate_ = player.playbackRate();
+  player.playbackRate(1);
+  if (player.controlBar.playbackRateMenuButton.playbackRateSupported &&
```
wseymour15, line 45: Will `player.control.playbackRateMenuButton` always be initialized here? If so, no change is needed

wseymour15 (approved):
Looks good! +1 after a response to my question.

src/adBreak.js
gkatsev: It's definitely worth doing null checks for the entire chain.

philjhale: Thanks. I'll add the null checks.

PR #551, opened 2023-05-02
feat: Add support for US Privacy macro for CCPA compliance

alex-barstow:
## Description
This PR adds support for a `{usp.uspString}` macro that is replaced with the US Privacy consent string, which is acquired using the IAB's [USP API](https://github.com/InteractiveAdvertisingBureau/USPrivacy/blob/master/CCPA/USP%20API.md) if it is available. The implementation follows the [reference code](https://github.com/InteractiveAdvertisingBureau/CCPA-reference-code) from IAB--  the API is invoked directly if it is available in the same window as the player, or via `postMessage()` if in a parent window. This also follows the same basic pattern as our implementation of TCF in this plugin, with a key difference. TCF triggers an event whenever the consent string changes, which lets us easily keep the string updated. The USP API does not, and manual asynchronous API calls are necessary to update it.

#### Test pages with reference USP API implementation:
- [in-page](https://solutions.brightcove.com/abarstow/CCPA/index-ads.html)
- [iframe](https://solutions.brightcove.com/abarstow/CCPA/index-ads-iframe.html)

Open the pages and select different options from the "Do Not Sell My Data" menu.

### DIFF
--- a/src/usPrivacy.js
+++ b/src/usPrivacy.js
@@ -1,0 +1,68 @@
+import window from 'global/window';
+
+const findUspApiLocatorWindow = (windowObj) => {
+  let targetWindow = windowObj.parent;
+
+  while (targetWindow !== windowObj.top) {
+    if (targetWindow.frames && targetWindow.frames.__uspapiLocator) {
+      return targetWindow;
+    }
+
+    targetWindow = targetWindow.parent;
+  }
+
+  return windowObj.top;
+};
+
+let uspString = '';
+
+export const getCurrentUspString = () => uspString;
+
+// Call the USP API to get the US Privacy String, either by invoking it directly or via postMessage() if inside an iframe.
+// In the former case the callback is synchronous, if the latter it is asynchronous, so to be safe it should always be
+// assumed to be asynchronous.
+// The window is passable as an argument for ease of testing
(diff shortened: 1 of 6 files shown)


src/usPrivacy.js
```diff
@@ -0,0 +1,68 @@
...
+        callId: uniqueId
+      }
+    };
+
+    const handleMessageEvent = (event) => {
+      if (
+        event &&
+        event.data &&
+        event.data.__uspapiReturn &&
+        event.data.__uspapiReturn.callId === uniqueId
```
roman-bc-dev, line 49: Random observation: I can't wait until we can start using optional chaining!

Essk (approved):

src/usPrivacy.js
mister-ben: Iirc last time we looked into that Babel replaced optional chaining in a very verbose way, increasing the dist versions size by a non-trivial amount.

gkatsev: The question is whether all the browsers that Video.js 8 supports has native optional chaining, in which case it may not compile it and we could use it.

mister-ben: Old Chrome is the problem. Introduced in v80.

gkatsev: huh, it's that new?

PR #552, opened 2023-05-05
feat: Add support for regex macros

alex-barstow:
## Description
This PR does two things:
- Adds support for replacing multiple macros that match a regular expression (see full description in readme changes)
- Simplifies and fixes a design flaw in the macro name override logic, specifically for pageVariable macros

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -59,21 +59,6 @@ const getDefaultValues = function(string) {
   return {defaults, modifiedString};
 };
 
-const processMacroNameOverrides = function(string, macroNameOverrides) {
-  const foundOverrides = {};
-
-  for (const defaultMacro in macroNameOverrides) {
-    const customMacro = macroNameOverrides[defaultMacro];
-
-    if (string.includes(customMacro)) {
-      foundOverrides[defaultMacro] = customMacro;
-      string = string.replace(new RegExp(customMacro, 'g'), defaultMacro);
-    }
-  }
-
-  return {string, foundOverrides};
-};
-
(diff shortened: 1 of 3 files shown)


src/macros.js
```diff
@@ -165,10 +158,22 @@ const getPageVariableMacros = function(string, defaults) {
...
     const resolvedMacroName = overrides.hasOwnProperty(macroName) ? overrides[macroName] : macroName;
 
-    string = string.split(resolvedMacroName).join(uriEncodeIfNeeded(macros[macroName], uriEncode));
+    if (resolvedMacroName.startsWith('r:')) {
+      try {
+        const regex = new RegExp(resolvedMacroName.slice(2), 'g');
+
+        string = string.replace(regex, uriEncodeIfNeeded(macros[macroName], uriEncode));
+      } catch (error) {
+        videojs.log.warn('Unable to replace regex macro. Please verify that you are passing valid regex.');
```
philjhale, line 170: I'm always in favour of making error messages as explicit and useful as possible. What is the effect of being unable to replace the regex macro? What won't work?

Also, is it worth outputting the offending regex?

Essk: I agree that outputting the input as seen by this function would be helpful to users who may well be programatically generating their regex strings.

Essk (commented):
This looks great overall, happy to approve once @philjhale's comment has been answered.

src/macros.js
alex-barstow: Thx, it now outputs the offending regex string

Essk (approved):

PR #553, opened 2023-05-18
fix: updateUsPrivacyString() callback not invoked when USP API not in use in any window

alex-barstow:
## Description
`player.ads.updateUsPrivacyString(callback)` would not invoke the callback if the USP API was not present in the current window or any parent window. Now it is synchronously invoked with a `null` privacyString, so `updateUsPrivacyString()` can be used safely when the presence of the USP API is uncertain.


wseymour15 (approved):
Looks good to me! Nice work!

adrums86 (approved):

roman-bc-dev (approved):

PR #565, opened 2024-01-10
fix: Toggle autoplay directly on tech when restoring snapshot for iOS

Essk:
Problem:  `restorePlayerSnapshot` uses `autoplay(true)` to force a content load in iOS Safari for correct placement of the playhead after an ad break. However if an integration has set `normalizeAutoplay: true` in the setup of `videojs`, `autoplay()` hands off to an async, programatic  play request with `manualAutoplay('play')`. The subsequent seek-back causes a visible restart of the content after ~1 second of playback. 

Solution: We can avoid the visual glitch by toggling autoplay directly on the HTML5 tech.  The existing followup mechanism to clear out the autoplay status continues to work as expected as `autoplay(false)` is passed to the tech anyway.

### DIFF
--- a/src/snapshot.js
+++ b/src/snapshot.js
@@ -222,9 +222,12 @@ export function restorePlayerSnapshot(player, callback) {
     player.one('contentloadedmetadata', restoreTracks);
 
     // adding autoplay guarantees that Safari will load the content so we can
-    // seek back to the correct time after ads
-    if (videojs.browser.IS_IOS && !player.autoplay()) {
-      player.autoplay(true);
+    // seek back to the correct time after ads.
+    // This is done directly on the html5 tech because if the integration has set
+    // normalizeAutoplay to true, the async play request via autoplay -> manualAutoplay_
+    // causes a visible skipback of the content after the ad break
+    if (videojs.browser.IS_IOS && !player.autoplay() && typeof player.techCall_ === 'function') {
+      player.techCall_('setAutoplay', true);
 
       // if we get here, the player was not originally configured to autoplay,
       // so we should remove it after it has served its purpose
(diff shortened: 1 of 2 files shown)


src/snapshot.js
```diff
@@ -222,9 +222,12 @@ export function restorePlayerSnapshot(player, callback) {
...
     // adding autoplay guarantees that Safari will load the content so we can
-    // seek back to the correct time after ads
-    if (videojs.browser.IS_IOS && !player.autoplay()) {
-      player.autoplay(true);
+    // seek back to the correct time after ads.
+    // This is done directly on the html5 tech because if the integration has set
+    // normalizeAutoplay to true, the async play request via autoplay -> manualAutoplay_
+    // causes a visible skipback of the content after the ad break
+    if (videojs.browser.IS_IOS && !player.autoplay() && typeof player.techCall_ === 'function') {
+      player.techCall_('setAutoplay', true);
```
misteroneill, line 230: So, if I understand correctly, setting it directly on the tech is synchronous and bypasses Video.js' built-in manual autoplay behavior?

misteroneill (commented):
Just want to make sure I understand correctly.

Is there a way to simulate the conditions that led to this in a test so we can catch any regressions?

src/snapshot.js
Essk: yes that's right.

A test or two would be good, I'll see if I can add something.

test/integration/test.snapshot.js
```diff
@@ -527,3 +527,64 @@ QUnit.test('Call play after live preroll on iOS', function(assert) {
...
+
+  this.player.ads.videoElementRecycled = () => true;
+  videojs.browser = {IS_IOS: true};
+  this.player.play = () => {};
+
+  this.player.trigger('loadstart');
+  this.player.trigger('adsready');
+  this.player.trigger('play');
+
+  assert.notOk(this.player.tech("it's fine").autoplay(), 'confirm autoplay on tech is off initially');
```
misteroneill, line 541: `it's fine` 😆

misteroneill (approved):
LGTM

PR #568, opened 2024-02-21
feat: ads error interface

wseymour15:
The implementation for an error interface specifically for ads. An error can be set on the `player.ads.error()` function by passing it an object containing a `type` string and a `metadata` object containing any custom data that is desired. The bulk of the work is in [ads.js](https://github.com/videojs/videojs-contrib-ads/pull/568/files#diff-4a89f8fbc0c2d6ec50e173833d52d84816b1d7e46c024dfa1c6de39a690c7f33)

There are examples of usage in this repo, but this can also be used when contrib ads is pulled into a user's repo. Errors can be handled or set by the user.

**Proposed Changes**:
- Error interface so we can listen for ad errors in other ares of the player.
- Trigger a handful of errors
- Test Changes
- Documentation Changes

### DIFF
--- a/docs/integrator/api.md
+++ b/docs/integrator/api.md
@@ -91,14 +91,14 @@ The error will be in the form of an object. This object must contain a `errorTyp
 
 To get the current ads error, the `error` function should be called with no parameters. Returns `null` if there is no error.
 ```js
-const currentAd = player.ads.error();
+const currentAdError = player.ads.error();
 ```
 
 To set the current ads error, the `error` function should be called with a valid error.
 ```js
-const ad = { errorType: 'ads-error-type' }
+const adError = { errorType: 'ads-error-type' }
 
-player.ads.error(ad);
+player.ads.error(adError);
 ```
 
 To clear the current ads error, the `error` function should be called with a `null` value.
--- a/docs/integrator/api.md
+++ b/docs/integrator/api.md
@@ -75,6 +75,38 @@ if (player.ads.inAdBreak()) {
 }
 ```
 
+### error
+
+Advanced function. Gets, sets, and clears a non-critical ads error in videojs-contrib-ads.
+
+The error will be in the form of an object. This object must contain a `errorType` string, and can also include zero to many custom properties.
+```js
+{
+  errorType: 'ads-error-type',
+  // The key can have any name, and the value can have any type of data.
+  customData: 'custom-data'
+  ...
+}
+```
+
+To get the current ads error, the `error` function should be called with no parameters. Returns `null` if there is no error.
+```js
+const currentAd = player.ads.error();
+```
+
+To set the current ads error, the `error` function should be called with a valid error.
+```js
+const ad = { errorType: 'ads-error-type' }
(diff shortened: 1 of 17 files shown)


src/ads.js
```diff
@@ -226,6 +231,49 @@ export default function getAds(player) {
...
+      if (err) {
+        this._error = null;
+
+        return;
+      }
+
+      this._error = err;
+
+      // TODO: log error
+      // do we want warn instead? Do we want to log here, or leave that up to when `error` is called?
```
wseymour15, line 269: I would appreciate feedback on thoughts whether we should log here or not.

docs/integrator/api.md
```diff
@@ -75,6 +75,38 @@ if (player.ads.inAdBreak()) {
...
+  errorType: 'ads-error-type',
+  // The key can have any name, and the value can have any type of data.
+  customData: 'custom-data'
+  ...
+}
+```
+
+To get the current ads error, the `error` function should be called with no parameters. Returns `null` if there is no error.
+```js
+const currentAd = player.ads.error();
```
dzianis-dashkevich, line 94:
```suggestion
const currentAdError = player.ads.error();
```

docs/integrator/api.md
```diff
@@ -75,6 +75,38 @@ if (player.ads.inAdBreak()) {
...
+```
+
+To set the current ads error, the `error` function should be called with a valid error.
+```js
+const ad = { errorType: 'ads-error-type' }
```
dzianis-dashkevich, line 99:
```suggestion
const adError = { errorType: 'ads-error-type' }
```

docs/integrator/api.md
```diff
@@ -75,6 +75,38 @@ if (player.ads.inAdBreak()) {
...
+
+player.ads.error(ad);
```
dzianis-dashkevich, line 101:
```suggestion
player.ads.error(adError);
```

src/macros.js
```diff
@@ -172,6 +172,11 @@ const replaceMacros = function(string, macros, uriEncode, overrides = {}) {
 
         string = string.replace(regex, uriEncodeIfNeeded(macros[macroName], uriEncode));
       } catch (error) {
+        player.ads.error({
+          errorType: videojs.Error.AdsMacroReplacementFailed,
+          macro: macroName
```
dzianis-dashkevich, line 177: I guess we can pass original error here

src/snapshot.js
```diff
@@ -196,6 +196,10 @@ export function restorePlayerSnapshot(player, callback) {
       try {
         resume();
       } catch (e) {
+        player.ads.error({
+          errorType: videojs.Error.AdsResumeContentFailed
```
dzianis-dashkevich, line 200: should we pass it here as well?

dzianis-dashkevich (approved):

src/snapshot.js
wseymour15: Good call! I updated this in the latest commit

PR #569, opened 2024-03-01
feat: outstream mode

usmanonazim:
# Description

An outstream player is used to show video ads exclusively and requires no regular content in between ad breaks. This PR enhances contrib-ads by introducing a new set of states to allow for 'outstream' players (ad only). 

### Changes introduced
- Three new states for players in 'outstream' mode (ad only).
**OutstreamPending -> OutstreamPlayback -> OutstreamDone**
- A new `playerMode` option that can be used to toggle `outstream` mode
- A 0ms mp4 video file is added as a source for the player if it is in outstream mode (to enable the ad plugin to request and play the ad tag)


src/plugin.js
```diff
@@ -217,7 +220,16 @@ const contribAdsPlugin = function(options) {
   // But first, cast to boolean.
   settings.stitchedAds = !!settings.stitchedAds;
 
-  if (settings.stitchedAds) {
+  if (settings.playerMode === 'outstream') {
```
usmanonazim, line 223: I'm not sure where the best place to set the player's src in contrib-ads is. I was thinking either here in the plugin initialisation or in the first Outstream state (OutstreamPending(

roman-bc-dev: Instead of keeping the full string here, it might be nice to start a constants file and import it. Just to keep things a bit more concise and legible.

roman-bc-dev: Also, to your point: since we're setting player instance listeners and parameters in this general block of code, I think it's fine to set the src here to consolidate if statements. I'd just add a comment for posterity.

src/states/OutstreamDone.js
```diff
@@ -0,0 +1,39 @@
...
+  static _getName() {
+    return 'OutstreamDone';
+  }
+
+  /*
+   * For state transitions to work correctly, initialization should
+   * happen here, not in a constructor.
+   */
+  init(player) {
+    player.ads._contentHasEnded = true;
```
roman-bc-dev, line 25: I understand that we may want this for consistency but is `_contentHasEnded` relevant for outstream players where there's no video content to scrub back through?

src/plugin.js
usmanonazim: Good point! Will create a constants file and put the video file in there :)

src/states/OutstreamDone.js
usmanonazim: That's true actually, I was doing it for consistency but it doesn't really make sense to have this - removing it now 😁

roman-bc-dev (approved):

